PR #13, opened 2018-01-13
Extract all supported functions.

gitlost:
See https://github.com/wp-cli/makepot-command/issues/12#issuecomment-357444799

Renames function map names to be more meaningful in `WordPress_Code_Extractor` and supports all in `WordPress_Functions_Scanner::saveGettextFunctions()` and errors if unmapped found.

Also errors if `comments_number_link()` found (not sure what to do here).

Adds tests.


gitlost:
https://github.com/oscarotero/Gettext requires PHP 5.4!

swissspidy:
Hmm. I simply copied `comments_number_link()` from the existing `makepot.php` file. Turns out, this isn't even an existing function in WordPress. At least not anymore.

It was added to the script over seven years ago in https://i18n.trac.wordpress.org/changeset/15325. I guess it's a remnant from the earlier days of WordPress.

We should remove it from this script as it serves no purpose.

swissspidy:
Actually, I think we can remove support for `_n_js()` and `_nx_js()` as well since they don't exist in WordPress.

These functions were added to `makepot.php` in https://i18n.trac.wordpress.org/changeset/20807 to test patches for the [#22229](https://core.trac.wordpress.org/ticket/22229) ticket in WordPress core. I used the same function in another patch for [#20491](https://core.trac.wordpress.org/ticket/20491).

We can remove them now and re-add them as soon as they actually exist. However, I don't think these functions will ever make it to core in that form

PR #16, opened 2018-01-22
Require PHP 5.4 for the makepot command

### LINKED ISSUE #14: Gettext library requires PHP 5.4+
As @gitlost had noted in https://github.com/wp-cli/makepot-command/pull/13#issuecomment-357454257 , the [Gettext](https://github.com/oscarotero/Gettext) library that the `makepot` command uses requires PHP 5.4+, which makes it unfit to be bundled with the current WP-CLI, with its PHP 5.3.29+ requirement.

Our current options:

* Fork the [`oscarotero/Gettext`](https://github.com/oscarotero/Gettext) library, make it PHP 5.3-compatible and submit PR to upstream (which is unlikely to be accepted)
* Fork the [`oscarotero/Gettext`](https://github.com/oscarotero/Gettext) library into our own org and maintain our own PHP 5.3-compatible branch
* Keep the `makepot` as a separate command for now and rediscuss bundling it with WP-CLI 2.0

@gitlost, @swissspidy thoughts?

swissspidy:
See #14.


schlessera:
Hmm, wait... that means installing WP-CLI over Composer would also break on PHP 5.3... We might need to reinvestigate how to deal with this.

swissspidy:
I just copied this from the server command. PHP 5.3 is still tested there as well.

Why would this prevent this installing WP-CLI via Composer though? I get that this would happen when we specify the PHP dependency in `composer.json`, but not with this change.

schlessera:
The problem is this setting in the library we're trying to pull in: https://github.com/oscarotero/Gettext/blob/master/composer.json#L21

So, Composer will not be able to compose this, even if we only conditionally use it.

swissspidy:
Yeah I realized that too after I saw the build log 🙂

I'd remove the PHP 5.3 test for now (I'm not keen on using `--ignore-platform-reqs` or similar) and then deal with the PHP requirement in https://github.com/wp-cli/wp-cli/issues/4554.

In the future commands could all specify their minimum PHP requirement in `composer.json` to make this more transparent.

schlessera:
Ok, sounds good. Let's target the next major with this command then.

PR #22, opened 2018-02-20
Update license copyright date to 2018

pdaalder:


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (C) 2011-2017 WP-CLI Development Group (https://github.com/wp-cli/language-command/contributors)
+Copyright (C) 2011-2018 WP-CLI Development Group (https://github.com/wp-cli/language-command/contributors)
```
schlessera: Should be `i18n-command`, instead of `language-command`.

PR #24, opened 2018-03-06
Don’t try to extract anything when there are no PHP files

swissspidy:
This is commonly the case when adding a child theme where only strings from `style.css` should get extracted but nothing else.

Without this change, `\Gettext\Extractors\Extractor::getFiles()` throws an `InvalidArgumentException` when being passed an empty array.

Now, the function is not being called in that scenario. Plus, any exceptions will be caught an printed by `WP_CLI::error()`.

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -28,9 +28,18 @@ Feature: Generate a POT file of a WordPress plugin
 
   Scenario: Does include file headers.
     When I run `wp scaffold plugin hello-world --plugin_name="Hello World" --plugin_author="John Doe" --plugin_author_uri="https://example.com" --plugin_uri="https://foo.example.com"`
+    Then the wp-content/plugins/hello-world directory should exist
+    And the wp-content/plugins/hello-world/hello-world.php file should exist
 
     When I run `wp i18n make-pot wp-content/plugins/hello-world wp-content/plugins/hello-world/languages/hello-world.pot`
     Then the wp-content/plugins/hello-world/languages/hello-world.pot file should exist
+    Then STDOUT should be:
+      """
+      Plugin file detected.
+      Success: POT file successfully generated!
+      """
+    And STDERR should be empty
+    And the wp-content/plugins/hello-world/languages/hello-world.pot file should exist
     And the wp-content/plugins/hello-world/languages/hello-world.pot file should contain:
       """
       msgid "Hello World"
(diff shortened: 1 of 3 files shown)


features/makepot.feature
```diff
@@ -26,6 +26,28 @@ Feature: Generate a POT file of a WordPress plugin
     And STDERR should be empty
     And the wp-content/plugins/hello-world/languages/hello-world.pot file should exist
 
+  Scenario: Does include file headers.
+    When I run `wp scaffold plugin hello-world --plugin_name="Hello World" --plugin_author="John Doe" --plugin_author_uri="https://example.com" --plugin_uri="https://foo.example.com"`
+
+    When I run `wp i18n make-pot wp-content/plugins/hello-world wp-content/plugins/hello-world/languages/hello-world.pot`
```
schlessera: Not checking STDOUT here?

src/WordPressCodeExtractor.php
```diff
@@ -91,7 +91,17 @@ public static function fromFile( $file, Translations $translations, array $optio
 	 */
 	public static function fromDirectory( $dir, Translations $translations, array $options = [] ) {
 		static::$dir = $dir;
-		static::fromFile( static::getFilesFromDirectory( $dir ), $translations, $options );
+
+		$files = static::getFilesFromDirectory( $dir );
+
+		try {
+			if ( ! empty( $files ) ) {
+				static::fromFile( $files, $translations, $options );
```
schlessera: Unrelated to this PR, but makes me wonder whether `public static function fromFile( $file, ... )` should not be named `public static fromFiles( $files, ... )` instead (plural vor name and argument), as it loops over multiple files.
Alternatively, the loop could be extracted out of `fromFile()` and put into the surrounding code. Might make more sense, conceptually. In that case, the above would turn from an `if` into a simple `foreach`.

swissspidy: I totally agree that it's not ideal.

Just note that this method implements `ExtractorInterface::fromFile()` where the doc block says `@param array|string $file A path of a file or files`. The abstract `Extractor` class it extends does the same, so I tried to keep the same behavior.

schlessera: Okay, let's keep it as is for now then.

features/makepot.feature
swissspidy: Can't hurt I guess. Added now.

Issue #7, opened 2017-12-15
Rethink domain handling


swissspidy:
Right now, the script only parses gettext calls where the text domain matches the plugin slug by setting `$this->translations->setDomain( $this->slug )`.

This matches WordPress.org behaviour, see https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/.

However, sometimes developers include third-party libraries in their plugins with different text domains and still expect those to be extracted.

Since I'd like to keep the current behaviour as the default, maybe we can add an `--ignore-domain` option to the script?

The code could look like this (simplified):

```php
if ( ! Utils\get_flag_value( $assoc_args, 'ignore-domain' ) {
    $this->translations->setDomain( $this->slug );
}
```

bradyvercher:
I'm pretty sure the current WP tools don't check the text domain at all when extracting strings. They all get included in the generated POT file.

With a plugin that looks like this:

```php
<?php
/**
 * Plugin Name: Test Plugin
 */

__( 'No Text Domain' )
__( 'Valid Text Domain', 'test-plugin' );
__( 'Library Text Domain', 'library' );
```

node-wp-i18n, which is based on a forked version of the WP tools, generates this POT file:

```pot
# Copyright (C) 2017 
# This file is distributed under the same license as the Test Plugin package.
msgid ""
msgstr ""
"Project-Id-Version: Test Plugin\n"
"Report-Msgid-Bugs-To: https://wordpress.org/support/plugin/test-plugin\n"
"POT-Creation-Date: 2017-12-15 16:34:09+00:00\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=utf-8\n"
"Content-Transfer-Encoding: 8bit\n"
"PO-Revision-Date: 2017-MO-DA HO:MI+ZONE\n"
"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
"Language-Team: LANGUAGE <LL@li.org>\n"
"X-Generator: node-wp-i18n 1.0.4\n"

#: test-plugin.php:6
msgid "No Text Domain"
msgstr ""

#: test-plugin.php:7
msgid "Valid Text Domain"
msgstr ""

#: test-plugin.php:8
msgid "Library Text Domain"
msgstr ""

#. Plugin Name of the plugin/theme
msgid "Test Plugin"
msgstr ""
```

And the WP-CLI command generates this POT file:

```pot
# Copyright (C) 2017 Test Plugin
[...]

PR #26, opened 2018-04-27
JavaScript String Extraction

### LINKED ISSUE #30: Add option to skip PHP or JS extraction altogether
If someone uses a babel plugin or similar for JS text extraction, there's no need to do that again in the WP-CLI command.

1. Use tool X for JS extraction
2. `wp i18n make-pot` with the option to skip JS extraction _and_ with the option to merge with existing strings.

Related: #26, #27, #29.

swissspidy:
More and more plugins rely heavily on JavaScript and the [JS I18N library](https://github.com/WordPress/packages/tree/master/packages/i18n) established by Gutenberg.

Since neither makepot.php nor this WP-CLI command support JavaScript, they had to build their own [Babel plugin](https://github.com/WordPress/packages/tree/master/packages/babel-plugin-makepot) to extract JavaScript files.

However, one cannot expect users to a) use Babel b) use two tools to generate a simple POT file.

So, why not support this in this little WP-CLI command here? :-)

Even though the [Gettext](https://github.com/oscarotero/Gettext) library we use for the PHP part has a [JS functions scanner](https://github.com/oscarotero/Gettext/blob/81c05cb213e8e4828db7aabd9dd363367ebca9f2/src/Utils/JsFunctionsScanner.php), it is very limited: no support for comments, no clear indication which ECMAScript standard is supported, no usage of an AST, which is really important with such a quickly evolving language.

I decided to use [Peast](https://github.com/mck89/peast) to traverse a script's AST, which was incredibly easy. We might even consider suggesting this for the Gettext library.

Still needs some tests, but so far it works fine.

Fixes #30.


aduth:
>However, one cannot expect users to a) use Babel

While it's not implied, just to note that a common point of confusion is that its implementation as a Babel plugin has something to do with an opt-in decision to use newer language syntax, when in fact it was largely a combination of (a) taking advantage of the [Babylon](https://github.com/babel/babel/tree/master/packages/babylon) parser which, being at the core of Babel's language transpilation, is assumed to be well-maintained and (b) to combine into the same transpilation pipeline _if_ one decides to use newer language features (as Gutenberg does, avoiding two separate parse steps).

aduth:
>b) use two tools to generate a simple POT file.

While it may be implemented as such in its current form, there's nothing which requires this to be a two-step process.

swissspidy:
@aduth Totally get the point. There's nothing wrong with that and I don't think we can get around that anyway.

Although [JSX support is planned](https://github.com/mck89/peast/issues/1#issuecomment-387429508) for Peast (the parser used in this PR), it makes more sense to use something like the Babel plugin for core and Gutenberg. That's why I've been working on things like #31 to make it easier to use both hand in hand.

As mentioned in yesterday's meeting, we'll likely test such a setup with Gutenberg and some other plugin(s) on WordPress.org soon. That way one at least won't have to generate these PHP files anymore during a build.

swissspidy:
Did some updates today:

* Fix merge conflicts
* Add support for JSX
* Add option to skip JS extraction altogether (see #30).  
  Useful when this is already done in another build step, e.g. through Babel.

src/MakePotCommand.php
```diff
@@ -140,7 +153,7 @@ protected function retrieve_main_file_data() {
 
 		$files = new IteratorIterator( new DirectoryIterator( $this->source ) );
 
-		/** @var DirectoryIterator $file */
+		/* @var DirectoryIterator $file */
```
schlessera: This should be a real docblock `/** */`, but inside of the `foreach` block as the first line, as `$file` only exists within the `foreach` block.
This will then allow an IDE to know how to handle `$file` in there.

src/MakePotCommand.php
```diff
@@ -73,8 +83,9 @@ class MakePotCommand extends WP_CLI_Command {
 	 * @when before_wp_load
 	 */
 	public function __invoke( $args, $assoc_args ) {
-		$this->source = realpath( $args[0] );
-		$this->slug   = Utils\get_flag_value( $assoc_args, 'slug', Utils\basename( $this->source ) );
+		$this->source  = realpath( $args[0] );
+		$this->slug    = Utils\get_flag_value( $assoc_args, 'slug', Utils\basename( $this->source ) );
+		$this->scan_js = ! Utils\get_flag_value( $assoc_args, 'skip-js', ! $this->scan_js );
```
schlessera: I'd prefer the property to be called `$skip_js`, to avoid the weird double negation here. As it is only used once, just directly mapping the parameter seems clearer to me.

src/JsFunctionsScanner.php
```diff
@@ -0,0 +1,176 @@
...
+namespace WP_CLI\I18n;
+
+use Gettext\Translations;
+use Gettext\Utils\JsFunctionsScanner as GettextJsFunctionsScanner;
+use Gettext\Utils\ParsedComment;
+use Peast\Peast;
+use Peast\Syntax\Node\Node;
+use Peast\Syntax\Node\VariableDeclaration;
+use Peast\Traverser;
+use Peast\Syntax\Node\CallExpression;
```
schlessera: All of these `Node\Something` imports are a case where I would prefer the use of relative namespaces.
So, have only 1 import `Peast\Syntax\Node` here, and then use relative namespaces like `Node\Identifier` or `Node\Literal` in the code.
This makes the import list less daunting, and it gives more context to the consuming code, where it is clear that these are all different types of nodes.

schlessera: Hmm, I now see that the library has a `Node\Node`, which is unfortunate and might make my above suggestion weird.

I usually build my code so that the interface is outside of the namespace with the implementations. This way, you can use `Node` as well as `Node\Literal`, ...

I'll let you decide what you find cleaner, but having a `Node\Node` inside the code is certainly not really elegant.

src/JsFunctionsScanner.php
```diff
@@ -0,0 +1,176 @@
...
+use Peast\Syntax\Node\Identifier;
+use Peast\Syntax\Node\Literal;
+
+class JsFunctionsScanner extends GettextJsFunctionsScanner {
```
schlessera: Class should be `final` and protected methods/properties should be private (unless they were already protected in the superclass).

src/PhpCodeExtractor.php
```diff
@@ -11,7 +11,7 @@
 use SplFileInfo;
 use WP_CLI;
 
-class WordPressCodeExtractor extends PhpCode {
+class PhpCodeExtractor extends PhpCode {
```
schlessera: Class should be `final` and protected methods/properties should be private (unless they were already protected in the superclass).

src/PhpFunctionsScanner.php
```diff
@@ -3,9 +3,9 @@
 namespace WP_CLI\I18n;
 
 use Gettext\Translations;
-use Gettext\Utils\PhpFunctionsScanner;
+use Gettext\Utils\PhpFunctionsScanner as GettextPhpFunctionsScanner;
 
-class WordPressFunctionsScanner extends PhpFunctionsScanner {
+class PhpFunctionsScanner extends GettextPhpFunctionsScanner {
```
schlessera: Class should be `final` and protected methods/properties should be private (unless they were already protected in the superclass).

src/JsCodeExtractor.php
```diff
@@ -0,0 +1,133 @@
...
+
+use Gettext\Extractors\JsCode;
+use Gettext\Translations;
+use Peast\Syntax\Exception as PeastException;
+use RecursiveCallbackFilterIterator;
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+use WP_CLI;
+
+class JsCodeExtractor extends JsCode {
```
schlessera: Class should be `final` and protected methods/properties should be private (unless they were already protected in the superclass).

schlessera: Generally, we don't want to add interfaces when this is not necessary. An interface is always a maintenance burden. As people can easily extend `JsCode` themselves, keeping `JsCodeExtractor` extensible is not needed.

src/JsFunctionsScanner.php
```diff
@@ -0,0 +1,176 @@
...
+				case 'single_plural_number_context_domain':
+					if ( ! isset( $args[4] ) ) {
+						break;
+					}
+
+					list( $original, $plural, $number, $context, $domain ) = $args;
+					break;
+			}
+
+			// Todo: Require a domain?
```
schlessera: Should we discuss this? Is there a technical difficulty? What is blocking this?

src/JsFunctionsScanner.php
```diff
@@ -0,0 +1,176 @@
...
+			if ( (string) $original !== '' && ( $domain === null || $domain === $translations->getDomain() ) ) {
+				$translation = $translations->insert( $context, $original, $plural );
+				$translation->addReference( $file, $node->getLocation()->getStart()->getLine() );
+
+				/* @var \Peast\Syntax\Node\Comment $comment */
```
schlessera: This one is not imported as the others are.

schlessera: This should be a real docblock inside of the `foreach` block.

src/JsFunctionsScanner.php
```diff
@@ -0,0 +1,176 @@
...
+				foreach( $argument->getLeadingComments() as $comment ) {
+					$all_comments[] = $comment;
+				}
+
+				if ( 'Identifier' === $argument->getType() ) {
+					$args[] = ''; // The value doesn't matter as it's unused.
+				}
+
+				if ( 'Literal' === $argument->getType() ) {
+					/* @var Literal $argument */
```
schlessera: This should be a real docblock.

src/JsFunctionsScanner.php
```diff
@@ -0,0 +1,176 @@
...
+			}
+
+			foreach( $callee->getLeadingComments() as $comment ) {
+				$all_comments[] = $comment;
+			}
+
+			$domain = $context = $original = $plural = null;
+			$args   = [];
+
+			/* @var Node $argument */
```
schlessera: This should be a real docblock inside of the `foreach` block.

src/JsFunctionsScanner.php
```diff
@@ -0,0 +1,176 @@
...
+				return;
+			}
+
+			/* @var CallExpression $node */
+			foreach ( $node->getArguments() as $argument ) {
+				// Support nested function calls.
+				$argument->setLeadingComments( $argument->getLeadingComments() + $node->getLeadingComments() );
+			}
+
+			/* @var Identifier $callee */
```
schlessera: This should be a real docblock.

src/JsFunctionsScanner.php
```diff
@@ -0,0 +1,176 @@
...
+			foreach( $node->getLeadingComments() as $comment ) {
+				$all_comments[] = $comment;
+			}
+
+			/* @var CallExpression $node */
+			if ( 'CallExpression' !== $node->getType() || 'Identifier' !== $node->getCallee()->getType() ) {
+				return;
+			}
+
+			/* @var CallExpression $node */
```
schlessera: This should be a real docblock inside of the `foreach` block.

src/JsFunctionsScanner.php
```diff
@@ -0,0 +1,176 @@
...
+		$traverser->addFunction( function ( $node ) use ( &$translations, $options, &$all_comments ) {
+			$functions = $options['functions'];
+			$file      = $options['file'];
+
+			/* @var Node $node */
+			foreach( $node->getLeadingComments() as $comment ) {
+				$all_comments[] = $comment;
+			}
+
+			/* @var CallExpression $node */
```
schlessera: This should be a real docblock.

src/JsFunctionsScanner.php
```diff
@@ -0,0 +1,176 @@
...
+		 * Traverse through JS code to find and extract gettext functions.
+		 *
+		 * Make sure translator comments in front of variable declarations
+		 * and inside nested call expressions are available when parsing the function call.
+		 */
+		$traverser->addFunction( function ( $node ) use ( &$translations, $options, &$all_comments ) {
+			$functions = $options['functions'];
+			$file      = $options['file'];
+
+			/* @var Node $node */
```
schlessera: This should be a real docblock inside of the `foreach` block.

src/JsCodeExtractor.php
```diff
@@ -0,0 +1,133 @@
...
+						return true;
+					}
+
+					return ( $file->isFile() && 'js' === $file->getExtension() );
+				}
+			),
+			RecursiveIteratorIterator::CHILD_FIRST
+		);
+
+		/* @var \DirectoryIterator $file */
```
schlessera: This should be a real docblock inside of the `foreach` block.

src/PhpCodeExtractor.php
```diff
@@ -118,42 +111,40 @@ public static function fromDirectory( $dir, Translations $translations, array $o
...
-				if ( $file->isFile() ) {
-					$filtered_files[] = $file->getPathname();
+
+					return ( $file->isFile() && 'php' === $file->getExtension() );
 				}
+			),
+			RecursiveIteratorIterator::CHILD_FIRST
+		);
+
+		/* @var \DirectoryIterator $file */
```
schlessera: This should be a real docblock inside of the `foreach` block.

schlessera: This should be imported instead of using a FQCN.

src/JsCodeExtractor.php
schlessera: This should be imported instead of using a FQCN.

src/MakePotCommand.php
swissspidy: @schlessera What IDE do you use? PhpStorm for example handles this just fine outside the `foreach`.

In WordPress core both `/**` and `/*` is used for these annotations. Not sure if WPCS would think the former is a badly formatted doc block.

src/JsCodeExtractor.php
swissspidy: One could also argue that this would result in an unused import statement as it's only used for the code comment.

Is this defined in the coding standards somewhere?

src/MakePotCommand.php
schlessera: I use PHPStorm as well, and for a few versions now, it interprets this correctly. But technically, the variable does not exist within the scope where you qualify it, and some other editors and static analysis tools will flag this.

I usually do one of these two instead:
```
foreach ( $files as $file ) {
	/** @var DirectoryIterator $file */
	if ( $file->isFile() && $file->isReadable() && 'php' === $file->getExtension()) {
		$plugin_files[] = $file->getRealPath();
	}
}
```
```
foreach ( $files as /** @var DirectoryIterator $file */ $file ) {
	if ( $file->isFile() && $file->isReadable() && 'php' === $file->getExtension()) {
		$plugin_files[] = $file->getRealPath();
	}
}
```

Regarding the docblock:
Although PHPStorm is generally very lax in this regard, using `@var` inside of a regular multiline comment does not have any semantic meaning whatsoever. Tags only have a proper meaning within docblocks. Again, some tools are more strict about this and will ignore the `@var` if it is not within a proper docblock.

src/JsFunctionsScanner.php
swissspidy: This is some copy/paste leftover from the original implementation, not something new. It's better read as `Todo: Should we really require a domain here?`.

Some background can be found in #7.

Basically, right now a text domain is required. It's not possible to not use one.

I now opened #35 to start a new discussion around this and will remove the comment in the mean time. It doesn't really communicate what the issue at hand is.

src/JsCodeExtractor.php
schlessera: The coding standards don't deal with namespaces, so there's no imports either.

It is not an unused import, as you're referencing it through the docblock. Note that PHP annotations like `@var` are considered parsable and executable code. So the annotation is something like `new var( DirectorIterator::class, 'file' )`.

src/MakePotCommand.php
schlessera: Hmm, nevermind my comments about moving the docblock inside of the `foreach`. This is apparently handled as an exception in the proposed PSR: https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md

image 2018-07-04 at 9 41 08 am

schlessera: However, the above standard proposal also states that tags need to be within a docblock, not a regular comment.

schlessera:
🎉 

We will probably still need to iterate on this in the future, but having it merged now means we can better test it in actual use cases.

PR #31, opened 2018-05-09
Add ability to merge with existing POT file

### LINKED ISSUE #29: Add option to merge newly extracted strings with existing ones
If there's already a POT file (either at the specified destination or one passed via an argument), e.g. one created by https://github.com/WordPress/packages/tree/master/packages/babel-plugin-makepot (see #26), it could be useful to merge these with the ones extracted by `make-pot`.

Example use case:

1. Extract JS strings using Babel, store them in `my-plugin.pot`.
2. Extract PHP strings using `wp i18n make-pot`, merge with the ones from `my-plugin.pot`, store both in `my-plugin.pot` again.

There could be a new option, `--merge`, which can be either used without value (= merge with destination file if it exists) or with a specified POT file (`--merge=my-plugin-js-strings.pot`).

This would help with point 2 mentioned in https://make.wordpress.org/core/2018/05/01/javascript-internationalization-the-missing-pieces/:

> Use that POT file to write the exact same internationalization functions in a “fake” PHP file that can be scanned by the WordPress.org translation platform. This will result in PO and MO files containing all of your plugin’s translations.

swissspidy:
Fixes #29.

Todo:

* [x] Make sure the headers of the resulting POT file aren't overridden by the existing one.
* [x] Figure out how to omit `--merge` and have it pick the destination file.

### DIFF
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -122,6 +122,8 @@ class MakePotCommand extends WP_CLI_Command {
 
 				$this->merge = $assoc_args['merge'];
 			}
+		}
+
 		if ( isset( $assoc_args['exclude'] ) ) {
 			$this->exclude = array_filter( array_merge( $this->exclude, explode( ',', $assoc_args['exclude'] ) ) );
 			$this->exclude = array_map(function($exclude) {
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -67,12 +67,12 @@ class MakePotCommand extends WP_CLI_Command {
 	 *
 	 * [--merge[=<file>]]
 	 * : Existing POT file file whose content should be merged with the extracted strings.
-	 * By default, the following files and folders are ignored: node_modules, .git, .svn, .CVS, .hg, vendor.
-	 * Leading and trailing slashes are ignored, i.e. `/my/directory/` is the same as `my/directory`.
+	 * If left empty, defaults to the destination POT file.
 	 *
 	 * [--exclude=<paths>]
 	 * : Include additional ignored paths as CSV (e.g. 'tests,bin,.github').
-	 * If left empty, defaults to the destination POT file.
+	 * By default, the following files and folders are ignored: node_modules, .git, .svn, .CVS, .hg, vendor.
+	 * Leading and trailing slashes are ignored, i.e. `/my/directory/` is the same as `my/directory`.
 	 *
 	 * ## EXAMPLES
 	 *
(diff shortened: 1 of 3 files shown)


schlessera:
I tried to look through the code to see what is missing for:
> Make sure the headers of the resulting POT file aren't overridden by the existing one

I wonder though what header data would be considered the correct one if you merge two files.

So, when using the `--merge` command, you can use two different paths:

1. Merge strings from the processed folder into the existing destination pot file.
2. Merge strings from the processed folder into an existing pot file, and create a separate destination file.

Here's what a general header would look like:
```
Project-Id-Version: My Plugin
Report-Msgid-Bugs-To: https://wordpress.org/support/plugin/my-plugin
Last-Translator: FULL NAME <EMAIL@ADDRESS>
Language-Team: LANGUAGE <LL@li.org>
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit
POT-Creation-Date: 2018-05-02T22:06:24+00:00
PO-Revision-Date: 2018-05-02T22:06:24+00:00
X-Domain: my-plugin
```

Here's my first guess at how headers would be handled then, provided that we can assume we're always working on the same "Project":

- `Project-Id-Version` & `X-Domain` should be required to be the same. Throw error if they are not.
- Use `POT-Creation-Date` from destination file (existing for 1., new for 2.)
- Override all other header settings

Does the above make sense?

src/MakePotCommand.php
```diff
@@ -53,18 +60,20 @@ class MakePotCommand extends WP_CLI_Command {
...
 	 *
 	 * [--domain=<domain>]
 	 * : Text domain to look for in the source code. Defaults to the plugin/theme slug.
 	 *
-	 * [--exclude=<paths>]
-	 * : Include additional ignored paths as CSV (e.g. 'tests,bin,.github').
-	 *
+	 * [--merge[=<file>]]
+	 * : Existing POT file file whose content should be merged with the extracted strings.
 	 * By default, the following files and folders are ignored: node_modules, .git, .svn, .CVS, .hg, vendor.
```
schlessera: Merging seems to have messed up the docblocks. `--merge` & `--exclude` have the wrong explanations.

schlessera: This will also need a refresh of the README.md once fixed.

swissspidy: Thanks, fixed now!

schlessera:
Awesome, @swissspidy, merged another one!

Issue #27, opened 2018-04-27
Add option to exclude certain folders and files


swissspidy:
The make-pot command currently scans all files with a `.php` extension (or `.js`, see #26).

In many cases, there might be files and folders that shouldn't be scanned. Some examples:

* node_modules/ (npm modules)
* Gruntfile.js
* webpack.config.js
* vendor/ (Composer dependencies)
* `.git`, `.svn`, `.svn` -> basically anything with a dot prefix

Being able to exclude these files would increase performance and create less POT files with potentially unwanted strings in it.

Perhaps there could be even a default list of folders and files to include.

PR #32, opened 2018-05-30
Exclude some common directories

swissspidy:
Does not yet contain a CLI argument to extend/override exclusions.

See #27.

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -740,5 +740,59 @@ Feature: Generate a POT file of a WordPress plugin
     When I run `wp i18n make-pot foo-plugin foo-plugin.pot --exclude="/myvendor/,/ignored.php"`
     Then the foo-plugin.pot file should not contain:
       """
+      I am being ignored
+      """
+
+  Scenario: Excludes nested folders and files
+    Given an empty foo-plugin directory
+    And a foo-plugin/foo-plugin.php file:
+      """
+      <?php
+      /**
+       * Plugin Name: Foo Plugin
+       * Plugin URI:  https://example.com
+       * Description:
+       * Version:     0.1.0
+       * Author:
+       * Author URI:
+       * License:     GPL-2.0+
+       * License URI: http://www.gnu.org/licenses/gpl-2.0.txt
+       * Text Domain: foo-plugin
+       * Domain Path: /languages
+       */
+       __( 'Hello World', 'foo-plugin' );
+      """
+    And a foo-plugin/some/sub/folder/foo.php file:
+      """
+      <?php
+       __( 'I am being ignored', 'foo-plugin' );
+      """
+    And a foo-plugin/some/other/sub/folder/foo.php file:
+      """
+      <?php
+       __( 'I am being ignored', 'foo-plugin' );
+      """
+    And a foo-plugin/bsome/sub/folder/foo.php file:
+      """
+      <?php
+       __( 'I am not being ignored', 'foo-plugin' );
+      """
+    And a foo-plugin/some/sub/folder.php file:
+      """
+      <?php
+       __( 'I am not being ignored either', 'foo-plugin' );
+      """
+
+    When I run `wp i18n make-pot foo-plugin foo-plugin.pot --exclude="some/sub/folder,other/sub/folder/foo.php"`
+    Then the foo-plugin.pot file should not contain:
+      """
+      I am being ignored
+      """
+    And the foo-plugin.pot file should contain:
+      """
       I am not being ignored
+      """
+    And the foo-plugin.pot file should contain:
+      """
+      I am not being ignored either
       """
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -523,3 +523,75 @@ Feature: Generate a POT file of a WordPress plugin
       """
       msgid "https://foobar.example.com"
       """
+
+  Scenario: Skips excluded folders
+    Given an empty foo-plugin directory
+    And a foo-plugin/foo-plugin.php file:
+      """
+      <?php
+      /**
+       * Plugin Name: Foo Plugin
+       * Plugin URI:  https://example.com
+       * Description:
+       * Version:     0.1.0
+       * Author:
+       * Author URI:
+       * License:     GPL-2.0+
+       * License URI: http://www.gnu.org/licenses/gpl-2.0.txt
+       * Text Domain: foo-plugin
+       * Domain Path: /languages
+       */
+       __( 'Hello World', 'foo-plugin' );
+      """
+    And a foo-plugin/vendor/ignored.php file:
+      """
+      <?php
+       __( 'I am being ignored', 'foo-plugin' );
+      """
+
+    When I run `wp i18n make-pot foo-plugin foo-plugin.pot`
+    And the foo-plugin.pot file should not contain:
+      """
+      I am being ignored
+      """
+
+  Scenario: Skips additionally excluded folders
+    Given an empty foo-plugin directory
+    And a foo-plugin/foo-plugin.php file:
+      """
+      <?php
+      /**
+       * Plugin Name: Foo Plugin
+       * Plugin URI:  https://example.com
+       * Description:
+       * Version:     0.1.0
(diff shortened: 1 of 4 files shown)


danielbachhuber:
> Does not yet contain a CLI argument to extend/override exclusions.

Should it be a CLI argument or an environment variable?

schlessera:
As this is meant to extract code from a WordPress plugin or theme, I wonder whether it wouldn't be preferable to use a WordPress filter to make the exclusions customizable from within a plugin/theme...

danielbachhuber:
Here's some prior art: https://github.com/wp-cli/find-command#using

```
[--include_ignored_paths=<paths>]
	Include additional ignored paths as CSV (e.g. '/sys-backup/,/temp/').
```

swissspidy:
I'm sticking with an extra argument for now.

Wondering though if I should rename `exclude` to `ignore` 🤷‍♂️

features/makepot.feature
```diff
@@ -523,3 +523,75 @@ Feature: Generate a POT file of a WordPress plugin
...
+       __( 'Hello World', 'foo-plugin' );
+      """
+    And a foo-plugin/vendor/ignored.php file:
+      """
+      <?php
+       __( 'I am being ignored', 'foo-plugin' );
+      """
+
+    When I run `wp i18n make-pot foo-plugin foo-plugin.pot`
+    And the foo-plugin.pot file should not contain:
```
schlessera: Should be `Then ...` for readability.

features/makepot.feature
```diff
@@ -523,3 +523,75 @@ Feature: Generate a POT file of a WordPress plugin
...
+       __( 'I am being ignored', 'foo-plugin' );
+      """
+    And a foo-plugin/bar/ignored.php file:
+      """
+      <?php
+       __( 'I am being ignored', 'foo-plugin' );
+      """
+
+    When I run `wp i18n make-pot foo-plugin foo-plugin.pot --exclude=foo,bar`
+    And the foo-plugin.pot file should not contain:
```
schlessera: Should be `Then ...` for readability.

src/WordPressCodeExtractor.php
```diff
@@ -109,21 +111,37 @@ public static function fromDirectory( $dir, Translations $translations, array $o
...
+					}
+				),
 				RecursiveIteratorIterator::CHILD_FIRST
 			);
 
-			/** @var \DirectoryIterator $file */
+			/** @var SplFileInfo $file */
 			foreach ( $files as $file ) {
-				if ( $file->isFile() && 'php' === $file->getExtension()) {
+				if (  $file->isFile() ) {
```
schlessera: CS: double space.

src/WordPressCodeExtractor.php
```diff
@@ -133,4 +151,5 @@ protected static function getFilesFromDirectory( $dir ) {
 
 		return $filtered_files;
 	}
+
```
schlessera: CS: Redundant new line.

src/WordPressCodeExtractor.php
```diff
@@ -109,21 +111,37 @@ public static function fromDirectory( $dir, Translations $translations, array $o
...
 		$filtered_files = [];
 
 		try {
 			$files = new RecursiveIteratorIterator(
-				new RecursiveDirectoryIterator( $dir, RecursiveDirectoryIterator::SKIP_DOTS ),
+				new RecursiveCallbackFilterIterator(
+					new RecursiveDirectoryIterator( $dir, RecursiveDirectoryIterator::SKIP_DOTS ),
+					function ( $file, $key, $iterator ) use ( $exclude ) {
+						/** @var SplFileInfo $file */
+						if ( in_array( $file->getBasename(), $exclude, true ) ) {
```
schlessera: This will fail if a path is returned with a trailing slash. Using trailing slashes is often done for avoiding partial matches. This either needs specific handling of trailing slashes, or an explicit mention in the argument documentation.

features/makepot.feature
```diff
@@ -523,3 +523,75 @@ Feature: Generate a POT file of a WordPress plugin
...
+       __( 'I am being ignored', 'foo-plugin' );
+      """
+
+    When I run `wp i18n make-pot foo-plugin foo-plugin.pot`
+    And the foo-plugin.pot file should not contain:
+      """
+      I am being ignored
+      """
+
+  Scenario: Skips additionally excluded folders
```
schlessera: Tests should cover these as well to make sure they work (and continue to do so):
- subfolder (deeper than in immediate root)
- file
- file in subfolder (deeper than in immediate root)
- partial match (NOT being excluded)

src/WordPressCodeExtractor.php
swissspidy: I don't think `getBasename()` ever returns a path with a trailing slash. The values in `$exclude` could have indeed have trailing slashes, yes.

I guess I can safely add a change to remove any slashes at the beginning and the end of the excluded directories/files and document this.

features/makepot.feature
swissspidy: Agreed. Will add these in a spare moment.

Also, I was just wondering whether glob-like patterns should be supported using something like `fnmatch`, but that's probably something for a future iteration.

features/makepot.feature
```diff
@@ -595,3 +595,150 @@ Feature: Generate a POT file of a WordPress plugin
       """
       I am being ignored
       """
+
+  Scenario: Skips excluded subfolders
```
schlessera: For this scenario, I was thinking more along the lines of adding an exclude that contains a subfolder, like this:
```
wp i18n make-pot foo-plugin foo-plugin.pot --exclude=some/sub/folder
```

I don't think testing the other way around (omitting a subfolder that was part of parent folder being excluded) is needed, as all files within a parent folder would have the same problem then.

src/WordPressCodeExtractor.php
```diff
@@ -128,6 +128,13 @@ function ( $file, $key, $iterator ) use ( $exclude ) {
 							return false;
 						}
 
+						// Check for more complex paths, e.g. /some/sub/folder.
+						foreach( $exclude as $path_or_file ) {
+							if ( substr( $file->getPathname(), - strlen( $path_or_file ) ) === $path_or_file ) {
```
swissspidy: There's probably a better way to do this, but I couldn't think of one right now.

src/WordPressCodeExtractor.php
```diff
@@ -109,21 +111,44 @@ public static function fromDirectory( $dir, Translations $translations, array $o
...
+					new RecursiveDirectoryIterator( $dir, RecursiveDirectoryIterator::SKIP_DOTS ),
+					function ( $file, $key, $iterator ) use ( $exclude ) {
+						/** @var SplFileInfo $file */
+						if ( in_array( $file->getBasename(), $exclude, true ) ) {
+							return false;
+						}
+
+						// Check for more complex paths, e.g. /some/sub/folder.
+						foreach( $exclude as $path_or_file ) {
+							if ( substr( $file->getPathname(), - strlen( '/' . $path_or_file ) ) === '/' . $path_or_file ) {
```
schlessera: You should use `substr_compare` instead as it is optimized towards the comparison:

```
$substring = '/' . $path_or_file;
if ( substr_compare( $file->getPathname(), $substring, - strlen( $substring ) ) === 0 ) { }
```

schlessera: Also, I'm wondering how safe this is when using multibyte strings. Off the top of my head, I can't think of a scenario where the comparison would fail to produce the correct result, but I'm not 100% sure. Do you think this might be an issue?

swissspidy: Good call.

I can't think of a scenario either _right now_, but you never know :-)

We could use `mb_ereg` in this case:

```php
if ( false !== mb_ereg( preg_quote( '/' . $path_or_file ) . '$', $file->getPathname() ) ) {
	return false;
}
```

Feels more readable anyway.

schlessera:
Merged now. Good work, @swissspidy!

Issue #36, opened 2018-07-04
Explore usefulness for WordPress core


swissspidy:
Right now the command requires a plugin or theme header to work. These allow setting the slug, text domain, etc. easily.

But what if we could use this for WordPress core / WordPress.org as well?

The current [makepot.php](https://develop.svn.wordpress.org/trunk/tools/i18n/makepot.php) script handles various special cases:

* Different meta data for the various projects
* Multiple POT files for WordPress admin, frontend, and timezones  
  Here's where the new `--exclude` option comes in handy
* There's no text domain, so this would be a good use case for an `--ignore-domain` argument (#35)
* No plugin or theme header that can be parsed

PR #38, opened 2018-07-06
Add option to extract strings with any text domain

### LINKED ISSUE #35: Add option to ignore the text domain
This could be called `--ignore-domain`.

See #7 for some background.

swissspidy:
This even allows extracting strings without any domain, which wasn't possible before.

That would be useful for projects like core (see #36) and private projects that aren't on WordPress.org.

Fixes #35.

### DIFF
--- a/src/JsFunctionsScanner.php
+++ b/src/JsFunctionsScanner.php
@@ -84,8 +84,8 @@ final class JsFunctionsScanner extends GettextJsFunctionsScanner {
 				$all_comments[] = $comment;
 			}
 
-			$domain = $context = $original = $plural = null;
-			$args   = [];
+			$context = $plural = null;
+			$args    = [];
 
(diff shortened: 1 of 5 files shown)


src/JsFunctionsScanner.php
```diff
@@ -106,39 +106,23 @@ public function saveGettextFunctions( Translations $translations, array $options
 			switch ( $functions[ $callee->getName() ] ) {
 				case 'text_domain':
 				case 'gettext':
-					if ( ! isset( $args[1] ) ) {
-						break;
-					}
-
-					list( $original, $domain ) = $args;
+					list( $original, $domain ) = array_pad( $args, 2, null );
```
schlessera: Ah, padding with `null` to avoid PHP errors when you're short on arguments. Nice one!

PR #42, opened 2018-07-10
Add `@when before_wp_load` to command namespace

### LINKED ISSUE #40: Require database when you run i18n
I was trying the package on my host where I have VVV as VM, so is not easy to share the same database.
When I execute `wp i18n` alert me that the mysql connection didn't worked.

If I execute `make-pot` instead this error doesn't appear and I can use it without problems.
I am not sure if this behaviour on first level command require a db connection and the others can ignore that but just to let know about this error.

swissspidy:
Before this PR:

```bash
$ wp i18n
Error: Error establishing a database connection. This either means that the username and password information in your `wp-config.php` file is incorrect or we can’t contact the database server at `localhost`. This could mean your host’s database server is down.
```

With this PR:

```bash
$ wp i18n
usage: wp i18n make-pot <source> [<destination>] [--slug=<slug>] [--domain=<domain>] [--merge[=<file>]] [--exclude=<paths>] [--skip-js]

See 'wp help i18n <command>' for more information on a specific command.
```

Fixes #40.

### DIFF
--- a/src/CommandNamespace.php
+++ b/src/CommandNamespace.php
@@ -12,6 +12,7 @@ use WP_CLI\Dispatcher\CommandNamespace as BaseCommandNamespace;
  *     # Create a POT file for the WordPress plugin/theme in the current directory
  *     $ wp i18n make-pot . languages/my-plugin.pot
  *
+ * @when before_wp_load
  */
 class CommandNamespace extends BaseCommandNamespace {
 }


swissspidy:
@schlessera No worries. To prevent accidental commits to master we could always protect the branch in the repository settings.

schlessera:
The `master` branch is already protected. I was just delusional, that must be it.

PR #43, opened 2018-07-10
Text domain header

### LINKED ISSUE #41: Detect the textdomain by plugin header and not folder
I am using locally a folder that is not the right slug of the plugin and seems that use that (that is the same of the plugin file). 
In my case the slug is defined in the root file in the headers but is not used by this command.

swissspidy:
Fixes #41.

Note: while working on this I uncovered a bug in the destination file determination.

It previously used `isset()` to check whether the `Domain Path` header was present. However, it is _always_ set, but can be empty.

This meant that the default value, `$this->destination = $this->slug . '.pot';` was never used. Also, that default value is wrong anyway as it doesn't generate the POT file in the plugin's directory, but the current working dir.

### DIFF
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -127,11 +127,11 @@ class MakePotCommand extends WP_CLI_Command {
 		}
 
 		// Determine destination.
-		$this->destination = $this->source . DIRECTORY_SEPARATOR . $this->slug . '.pot';
+		$this->destination = "{$this->source}/{$this->slug}.pot";
 
 		if ( ! empty( $file_data['Domain Path'] ) ) {
 			// Domain Path inside source folder.
-			$this->destination = $this->source . DIRECTORY_SEPARATOR . $file_data['Domain Path'] . DIRECTORY_SEPARATOR . $this->slug . '.pot';
+			$this->destination = sprintf( '%s/%s/%s.pot', $this->source, $file_data['Domain Path'], $this->slug );
 		}
 
 		if ( isset( $args[1] ) ) {
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -127,18 +127,15 @@ class MakePotCommand extends WP_CLI_Command {
 		}
 
 		// Determine destination.
+		$this->destination = $this->source . DIRECTORY_SEPARATOR . $this->slug . '.pot';
+
+		if ( ! empty( $file_data['Domain Path'] ) ) {
+			// Domain Path inside source folder.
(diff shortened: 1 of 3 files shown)


src/MakePotCommand.php
```diff
@@ -106,29 +108,34 @@ public function __invoke( $args, $assoc_args ) {
...
-				$this->destination = $this->source . DIRECTORY_SEPARATOR . $file_data['Domain Path'] . DIRECTORY_SEPARATOR . $this->slug . '.pot';
+			if ( ! empty( $file_data['Text Domain'] ) ) {
+				$this->domain = $file_data['Text Domain'];
 			}
+
+			$this->domain = Utils\get_flag_value( $assoc_args, 'domain', $this->domain );
+		}
+
+		// Determine destination.
+		$this->destination = $this->source . DIRECTORY_SEPARATOR . $this->slug . '.pot';
```
schlessera: The use of `DIRECTORY_SEPARATOR` is actually not a good practice here. That constant should be sued for functions like `explode()` or `(r|l)trim`. If you compose a filesystem string for the server, just use `/` instead.

So, I'd write this line as follows:
```
$this->destination = "{$this->source}/{$this->slug}.pot";
```

src/MakePotCommand.php
```diff
@@ -106,29 +108,34 @@ public function __invoke( $args, $assoc_args ) {
...
+
+		if ( ! empty( $file_data['Domain Path'] ) ) {
+			// Domain Path inside source folder.
+			$this->destination = $this->source . DIRECTORY_SEPARATOR . $file_data['Domain Path'] . DIRECTORY_SEPARATOR . $this->slug . '.pot';
```
schlessera: Again, use of `DIRECTORY_SEPARATOR` counterproductive here.

src/MakePotCommand.php
```diff
@@ -165,6 +170,16 @@ public function __invoke( $args, $assoc_args ) {
...
 	}
 
+	/**
+	 * Removes leading and trailing slashes of a string.
+	 *
+	 * @param string $string What to add the remove slashes from.
+	 * @return string String without leading and trailing slashes.
+	 */
+	protected function unslashit( $string ) {
+		return ltrim( rtrim( $string, '/\\' ), '/\\' );
```
schlessera: This would actually be a good use of `DIRECTORY_SEPARATOR`. However, as we are not quite sure what we get and what we already did with it, I'd say we leave it as is now to stay on the safe side (especially with Cygwin on Windows).

src/MakePotCommand.php
swissspidy: Thanks, that makes sense!

Issue #50, opened 2018-07-20
conserve X-Poedit-* headers with --merge


drzraf:
When using: `wp i18n make-pot . languages/theme.pot --merge` on a pot file already created with POedit, I'd like to keep custom headers since they contain many useful hints for contributors (possibly using poedit).
Eg:
```
"X-Poedit-Basepath: ..\n"
"X-Poedit-KeywordsList: __;_e;_n;_x;_ex;_nx;esc_attr__;esc_attr_e;esc_attr_x;esc_html__;esc_html_e;esc_html_x;_n_noop;_nx_noop;translate_nooped_plural\n"
"Plural-Forms: nplurals=2; plural=(n != 1);\n"
"X-Poedit-SearchPath-0: .\n"
"X-Poedit-SearchPathExcluded-0: foo\n"
"X-Poedit-SearchPathExcluded-1: bar.php\n"
```

schlessera:
Yes, this makes sense to me.

I'll defer the review of the PR to @swissspidy, as he's the expert in that domain.

swissspidy:
Related Twitter thread: https://twitter.com/Mte90Net/status/1014512091036897281

I'm curious to know what exactly Poedit does with these headers. Are they really useful if you don't need to generate a POT file using Poedit, but only need to translate?

If so, we should correctly add these headers on our own, instead of just copying the ones from an existing POT file.

In your scenario, the POT file created by WP-CLI is the single source of truth, as it contains all translatable strings. If anything, WP-CLI should declare which paths were excluded (or perhaps combine that header), what the base path is, and which functions are scanned.

Also, the "Plural-Forms" header doesn't make for a POT file as it contains no language information and thus no plural form information.

drzraf:
There are useful since they preload the configuration used by Poedit when it scans project's directories.
Avoiding repetitive, error-prone, **re**configuration each time regenerated po files are loaded by Poedit is useful.

swissspidy:
Why do you need Poedit to scan the project directory when you already use WP-CLI to do that and create a POT file?

When you create a POT file using WP-CLI, all you need to do is create a new translation in Poedit based on that catalog. I don't see how Poedit needs these headers in that case.

drzraf:
Because pot file "may" (will) be used by people relying on Poedit for
some tasks.
There are no uniform way to manage collaborative translationsand
hard to impose a workflow (especially for new project). It's preferible
to not lock any door (including poedit)...
[...]

PR #51, opened 2018-07-20
preserve previous pot file headers when --merge is used

drzraf:
see #50


swissspidy:
Closing since #50 is a _wontfix_. Best case, these headers should be newly added, not copied from some other POT files. Also, existing POT files can have way more headers than just the Poedit ones anyway. I don't want to blindly add those.

PR #54, opened 2018-07-24
Factor directory iterator as a trait

drzraf:
Factor directory iterator as a trait so that other extractor can rely on it and make extension generic so that test (`<extension>` === 'php') can be overridden if needed.

### DIFF
--- a/src/IterableCodeExtractor.php
+++ b/src/IterableCodeExtractor.php
@@ -1,0 +1,122 @@
+<?php
+
+namespace WP_CLI\I18n;
+
+use Gettext\Translation;
+use Gettext\Translations;
+use RecursiveCallbackFilterIterator;
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+use DirectoryIterator;
+use WP_CLI;
+
+trait IterableCodeExtractor {
+
+	private static $dir = '';
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public static function fromFile( $file, Translations $translations, array $options = [] ) {
+		foreach ( static::getFiles( $file ) as $f ) {
+			// Make sure a relative file path is added as a comment.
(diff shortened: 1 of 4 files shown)


swissspidy:
That would need the same change for the JS extractor to make things DRY.

src/IterableCodeExtractor.php
```diff
@@ -0,0 +1,122 @@
...
+use RecursiveIteratorIterator;
+use DirectoryIterator;
+use WP_CLI;
+
+trait IterableCodeExtractor {
+
+	private static $dir = '';
+
+	/**
+	 * {@inheritdoc}
```
swissspidy: I'd use the docs from the [`ExtractorInterface`](https://github.com/oscarotero/Gettext/blob/cae84aff39a87e07bd6e5cddb5adb720a0ffa357/src/Extractors/ExtractorInterface.php) here as there's nothing to inherit from otherwise.

Plus, we could describe the possible options a bit better. Perhaps the way WordPress core does it: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-1-parameters-that-are-arrays

src/IterableCodeExtractor.php
```diff
@@ -0,0 +1,122 @@
...
+			static::fromString( $string, $translations, $options );
+		}
+	}
+
+	/**
+	 * Recursively extracts the translations from a directory.
+	 *
+	 * @param string $dir Root path to start the recursive traversal in.
+	 * @param Translations $translations The translations instance to append the new translations.
+	 * @param array $options
```
swissspidy: As noted above, we could describe the possible options a bit better. Perhaps the way WordPress core does it: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-1-parameters-that-are-arrays

schlessera:
Thanks for the pull-request, @drzraf !

Issue #46, opened 2018-07-10
Inconsistent debug messages


swissspidy:
When running `wp i18n make-pot` with the `--debug` flag, one currently sees messages like this:

```
Debug (make-pot): Plugin file: /foo-plugin/foo-plugin.php (0.089s)
Debug (make-pot): Destination: /foo-plugin//languages/glossary-by-codeat.pot (0.09s)
```

Besides fixing the wrong `//` in the output there, I think we could improve this output a bit more. For example, we could print the final text domain that will be used for scanning, and some other helpful details like the total number of found strings.

swissspidy:
Fixed by #56.

PR #56, opened 2018-07-26
Print some more helpful debug messages

swissspidy:
See #46.

### DIFF
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -319,7 +319,7 @@ class MakePotCommand extends WP_CLI_Command {
 
 		// Add existing strings first but don't keep headers.
 		if ( $this->merge ) {
-			WP_CLI::debug( sprintf( 'Mergining with existing POT file: %s', $this->merge ), 'make-pot' );
+			WP_CLI::debug( sprintf( 'Merging with existing POT file: %s', $this->merge ), 'make-pot' );
 
 			$existing_translations = new Translations();
 			Po::fromFile( $this->merge, $existing_translations );
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -241,7 +241,6 @@ class MakePotCommand extends WP_CLI_Command {
 
 		/** @var DirectoryIterator $file */
 		foreach ( $files as $file ) {
-			WP_CLI::debug($file->getRealPath());
 			if ( $file->isFile() && $file->isReadable() && 'php' === $file->getExtension()) {
 				$plugin_files[] = $file->getRealPath();
 			}
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -190,7 +190,7 @@ class MakePotCommand extends WP_CLI_Command {
 			}
 
 			if ( isset( $this->merge ) && ! file_exists( $this->merge ) ) {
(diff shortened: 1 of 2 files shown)


src/MakePotCommand.php
```diff
@@ -300,6 +319,8 @@ protected function makepot() {
 
 		// Add existing strings first but don't keep headers.
 		if ( $this->merge ) {
+			WP_CLI::debug( sprintf( 'Mergining with existing POT file: %s', $this->merge ), 'make-pot' );
```
schlessera: There's a typo here: `Mergining` => `Merging`

src/MakePotCommand.php
```diff
@@ -147,38 +151,50 @@ public function handle_arguments( $args, $assoc_args ) {
...
-					WP_CLI::error( sprintf( 'Invalid file %s', $assoc_args['merge'] ) );
-				}
-
 				$this->merge = $assoc_args['merge'];
 			}
-		}
 
-		WP_CLI::debug( sprintf( 'Destination: %s', $this->destination ), 'make-pot' );
+			if ( isset( $this->merge ) && ! file_exists( $this->merge ) ) {
+				WP_CLI::warning( sprintf( 'POT file to merge with does not exist: %s', $this->merge ) );
```
schlessera: Typo: `with` => `which`

src/MakePotCommand.php
```diff
@@ -377,7 +398,17 @@ protected function makepot() {
 			}
 		}
 
-		return PotGenerator::toFile( $this->translations, $this->destination );
+		$result = PotGenerator::toFile( $this->translations, $this->destination );
+
+		$translations_count = count( $this->translations );
+
+		if ( 1 === $translations_count ) {
+			WP_CLI::debug( sprintf( 'Extracted %d string', count( $this->translations ) ), 'make-pot' );
```
schlessera: Reuse `$translations_count` here

src/MakePotCommand.php
```diff
@@ -377,7 +398,17 @@ protected function makepot() {
...
+		} else {
+			WP_CLI::debug( sprintf( 'Extracted %d strings', count( $this->translations ) ), 'make-pot' );
```
schlessera: Reuse `$translations_count` here

src/MakePotCommand.php
schlessera: Ah, I read it in a different way: "Here is a reference to the POT file we wanted to merge but _which_ does not exist: ...".

Some random ideas for an alternative...

* "Merge target does not exist: ..."
* "POT file to be merged does not exist: ..."
* "Invalid POT file set as merge target: ..."
* "File to merge is not valid: ..."
* "Invalid file provided to --merge: ..."

src/MakePotCommand.php
schlessera: Maybe we should move over the `Inflector` class from `wp-cli/scaffold-package` for strings like this.
You could then do something along these lines:
```
WP_CLI::debug( sprintf(
   'Extracted %d %s',
   $translations_count,
   Utils\maybe_pluralize( 'string', $translations_count )
), 'make-pot' );
```

schlessera: See https://github.com/wp-cli/wp-cli/issues/4880

swissspidy: Only now saw this. The same applies to the language command as well, so 👍

Issue #52, opened 2018-07-20
wp-cli make-pot Twig (Timber) support


drzraf:
It would be nice for `wp-cli make-pot` to support twig template out of the box.
Some of the building blocks:
https://github.com/timber/timber/pull/1753
https://github.com/umpirsky/Twig-Gettext-Extractor/pull/60

swissspidy:
I just read through these issues as well as https://github.com/timber/timber/issues/1754 and https://timber.github.io/docs/guides/internationalization/.

Is Twig so popular in WordPress space and is Timber the de facto standard for it? Where would we draw the line when someone else comes with their templating engine and wants this included in `wp i18n make-pot`?

I am not really sure whether this is something WP-CLI should handle.

The Gettext library we use already supports Twig extraction:

https://github.com/oscarotero/Gettext/blob/eea87605224d25125a39a83fbc9bc1c28d16bb7a/src/Extractors/Twig.php
https://github.com/oscarotero/Gettext/blob/master/tests/assets/twig/input.php

If we can easily use that to make it work with Timber, I'm open for looking at it. Otherwise I'm not sure it's worth the effort.

Also, good to keep in mind that with this we'd need to scan (not necessarily load) every PHP file twice, once for good old gettext and once for Twig.

drzraf:
I'll try to offer a PR, but I hope bootstrap'ing WP is not a showstopper...
It's needed to:
1. load twig and its filters
2. load custom twig filters/functions from functions.php (if any)

swissspidy:
You mean one common POT file?

Let's say your custom command is `wp timber make-pot` which creates a new file called `twig.pot`.

Then, you can run `wp i18n make-pot --merge=twig.pot` to create one single POT file containing all translations from Twig, regular PHP code and JavaScript code.

Of course you could combine these commands, e.g. `wp timber make-pot && wp i18n make-pot --merge=twig.pot`.

swissspidy:
> (I wondered whether `make-pot --twig` could have be run after_wp_load
while keep `make-pot` before_wp_load).

I wondered the same yesterday, but I'm not sure it's possible. Also, it could lead to unexpected results for users and would need thorough documentation.

> It'd be annoying not being able to reuse the hundreds of useful lines already in
that package.

That's why I suggested extending this command. I bet it's possible to extend the `MakePotCommand` class in your own code.

[...]

PR #59, opened 2018-07-27
Functions scanner for Twig AST

drzraf:
I would have preferred to submit this to upstream oscarotero/Gettext but the way i18n function wrappers work make it a bit difficult (It's easier to inherit from `wp-cli` `PhpFunctionsScanner`)

Note: This extractor work at a "higher" level than current one (which usually run a PHP extractor after a Twig-compilation pass).


schlessera:
I'm open to removing the `final` keyword from `PhpFunctionsScanner` for now.

However, as already stated in https://github.com/wp-cli/i18n-command/issues/52#issuecomment-407689193, I don't think Twig support is a good fit for bundling right now.

Also, please note that we don't accept code that comes with individual copyright or changes licensing.

drzraf:
* `final` keyword: #62 
* about license, I wrongly believed that `wp-cli` was GPL licenced excepted this package (which relies upon an MIT-licensed lib'). Dual-license header removed (I hope in the future `wp-cli` could be GPL-ed, if not dual-licenced btw :))
* About Twig support. I tried to formalize the concerns in https://github.com/wp-cli/i18n-command/issues/52#issuecomment-408524211 (the main one was that Twig support would complexify command-line, especially argument handling/bootstrap level). I'm still trying to solve all the issues resulting from keeping this Twig support outside `i18n` package. But if it's not possible, I still believe that including the class wouldn't harm the package either: it'd just make one public class available to external commands (and it'd be logical since scanner/extractor rely more on i18n package than on anything else and was written so that it share the same logic).

drzraf:
I updated https://github.com/wp-cli/i18n-command/issues/52#issuecomment-408524211.
You will find there the limitations that keep me from inheriting make-pot `MakePotCommand` from a theme-defined subcommand.

drzraf:
@oscarotero: would it fit better [there](https://github.com/oscarotero/Gettext)?

oscarotero:
@drzraf I'm not a twig user so don't get the difference between this extractor and the one currently in Gettext.

drzraf:
### Gettext current one:
1. Transforms Twig as PHP code (= `tokenize()+parse()+render()`
2. Uses a PhpCode extractor (and its specific configuration about function name)
#### advantages:
* Ability to reuse PhpCode extractor to extract function call
#### disadvantage:
* Twig rendered php code can be transformed/wrapped thus hiding gettext functions. See https://github.com/timber/timber/pull/1753. In that case, PhpCode won't find the calls to `__()`, `gettext()`, ... because they are wrapped inside `call_user_func()`, ...
* Can't currently handle templates making use of custom Twig extensions.

### Proposed one:
1. Parse Twig generated AST tree (= `tokenize()+parse()`)
2. Recursively iterate over the tree to find function calls.
#### advantages:
* Narrower to Twig syntax. Eg: `{{ __("foo", "domain") }}`. AST has undergone less transformations so function wrapping isn't problem anymore
* More robust because it iterates over the AST from Twig parser, instead of looking at source-code
* Probably more efficient
* Since it takes an already initialized `$twig` environment in the constructor instead of a string, it **supports user-defined extensions** ¹
#### disadvantages:
* Can't rely upon PhpCode extractor anymore
* Configuration of `gettext()` function names to extract (`__`, `_x`, `_e`, `ngettext`, ...) needs to be known sooner than currently = at the functionsScanner level step.


¹ It's simple (and thus common) to extend Twig with additional filters/functions. (That's what Timber does for handling Twig i18n the-wordpress-way). But all these filters and functions must be defined *before* parsing the template. A good scanner+extractor must let user hook in the process to register their extensions on the Twig parser.

oscarotero:
@drzraf Ok, it looks fine to me.

I don't care about don't reusing PhpCode. It was created originally to scan php code, not to reuse by other template based extractors. I only have this suggestions:

- In this PR, you extended `PhpFunctionScanner` but without use any of the inherited methods, so it's useless. Due this code is specific for Twig and won't be used by other extractors, I'd put it directly in the [Twig extractor](https://github.com/oscarotero/Gettext/blob/master/src/Extractors/Twig.php)
- In the extractor you can pass the initialized `$twig` and the functions names in the $options argument.

If you want to work in a pull request, I'd appreciate that. 
Thanks!

drzraf:
FTR: PR at https://github.com/oscarotero/Gettext/pull/195

PR #63, opened 2018-07-31
separate translation extraction from Po file writing

drzraf:
split translation extraction from po file writing.
That way child class (or code instantiating this class) can do post processing on `$translations` array or append new translation (possibly fetched with different extractors), but intended to be stored in the same destination po file


src/MakePotCommand.php
```diff
@@ -295,7 +298,7 @@ protected function get_main_file_data() {
 	 *
 	 * @return bool True on success, false otherwise.
```
swissspidy: That's now incorrect with this change.

Instead of making `translations` a property of the `MakePotCommand`, I think we can just use `$translations = new Translations()` and return that object at the end (like you did already).

We could get rid of `set_default_headers()` and move these ~10 lines into `makepot()`.

Also, since `makepot()` doesn't really create a POT file anymore, we could call it `extract_strings()` or something.

src/MakePotCommand.php
```diff
@@ -111,10 +111,13 @@ class MakePotCommand extends WP_CLI_Command {
 	 */
 	public function __invoke( $args, $assoc_args ) {
 		$this->handle_arguments( $args, $assoc_args );
-		if ( ! $this->makepot() ) {
+		$this->makepot();
+		if ( !$this->translations ) {
```
swissspidy: Can we keep the space after the `!`? 🙂

src/MakePotCommand.php
```diff
@@ -111,10 +111,13 @@ class MakePotCommand extends WP_CLI_Command {
...
+			WP_CLI::warning( 'No translation found' );
+		}
+		if ( !PotGenerator::toFile( $this->translations, $this->destination ) ) {
```
swissspidy: Can we keep the space after the `!`? 🙂

swissspidy:
Thanks for this PR, @drzraf! I like where this is heading.

src/MakePotCommand.php
```diff
@@ -293,28 +291,39 @@ protected function get_main_file_data() {
 	/**
 	 * Creates a POT file and stores it on disk.
 	 *
-	 * @return bool True on success, false otherwise.
+	 * @return Translations A Translation set.
 	 */
-	protected function makepot() {
-		$this->translations = new Translations();
+	public function extract_strings() {
```
swissspidy: Note that every public method on a class is registered as a subcommand of the command. See https://make.wordpress.org/cli/handbook/commands-cookbook/#required-registration-arguments

I don't think we want this here :-) — `protected` should work fine for the sake of extensibility.-

drzraf: Ouch....
The workaround expressed in https://github.com/wp-cli/i18n-command/issues/52#issuecomment-409236464 (_attempt 4_) relies on the fact that instead of inheriting the command, I would be able to instantiate it and grab `$translations` by calling the [public] `extract_strings()` method.

swissspidy: PHPDoc is used for documenting a command and perhaps making aliases. Registration is done via `WP_CLI::add_command()`.

drzraf: => method set to protected

src/MakePotCommand.php
```diff
@@ -293,28 +291,39 @@ protected function get_main_file_data() {
...
+			$translations->mergeWith( $existing_translations, Merge::ADD | Merge::REMOVE );
 		}
 
 		$meta = $this->get_meta_data();
 		PotGenerator::setCommentBeforeHeaders( $meta['comments'] );
 
-		$this->set_default_headers();
+		$meta = $this->get_meta_data();
+
+		$translations->setHeader( 'Project-Id-Version', $meta['name'] . ' ' . $meta['version'] );
```
schlessera: Was there a specific reason why this block was inlined instead of remaining in a separate method? I'd consider this is a regression in terms of code quality.

swissspidy: I suggested it in an inline comment. I wanted to some refactoring in another PR after this one.

drzraf: Because it was requested^^
I restored the function.

schlessera: Ah, sorry @swissspidy, I missed that comment.

swissspidy:
Unfortunately there's now a minor merge conflict because of another recent PR.

@drzraf Would you mind merging `master` into the branch to resolve these?

Afterwards I'll happily merge this PR :)

Thanks in advance!

drzraf:
rebased + squashed

Issue #61, opened 2018-07-29
Add warnings for more scenarios


swissspidy:
Previously: #33.

-----

The WordPress Coding Standards project has a few sniffs that we could borrow from to improve the warnings created by the make-pot command.

These sniffs currently do:

* Check for the presence of a translators comment if one of the text strings contains a placeholder.
* Check for inconsistencies between single and plural arguments.
* Check for multiple unordered placeholders.
* Check that the translatable string itself is not empty

Source: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/eccb1f721f95fdcf5c2e383340aa4b094c0317ad/WordPress/Sniffs/WP/I18nSniff.php#L36-L59

Let's determine whether it makes sense to port some of these over to WP-CLI.

swissspidy:
Fixed by #64.

PR #64, opened 2018-07-31
Check for more mistakes in translatable strings

swissspidy:
See #61.


features/makepot.feature
```diff
@@ -620,7 +617,114 @@ Feature: Generate a POT file of a WordPress project
...
+    Given an empty foo-plugin directory
+    And a foo-plugin/foo-plugin.php file:
+      """
+      <?php
+      /**
+       * Plugin Name: Plugin name
+       */
+
+      sprintf(
+        _n( '%1$s Comment (%2$s)', '%2$$s Comments (%1$s)', $number, 'foo-plugin' ),
```
schlessera: double dollar sign here: `%2$$s`

src/MakePotCommand.php
```diff
@@ -381,34 +427,99 @@ protected function makepot() {
...
-			WP_CLI::debug( sprintf( 'Extracted %d strings', $translations_count ), 'make-pot' );
+			if ( $translation->hasPlural() ) {
+				preg_match_all( self::SPRINTF_PLACEHOLDER_REGEX, $translation->getOriginal(), $single_placeholders );
+				$single_placeholders = $single_placeholders[0];
+
+				preg_match_all( self::SPRINTF_PLACEHOLDER_REGEX, $translation->getPlural(), $plural_placeholders );
+				$plural_placeholders = $plural_placeholders[0];
+
+				// see https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#plurals
+				if ( count( $single_placeholders ) < \count( $plural_placeholders ) ) {
```
schlessera: Two different forms of `count()` used, one relative and one absolute.

src/MakePotCommand.php
```diff
@@ -381,34 +427,99 @@ protected function makepot() {
...
+					WP_CLI::warning( sprintf(
+						'Missing singular placeholder, needed for some languages. See https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#plurals %s',
+						$location
+					) );
+				} else {
+					sort( $single_placeholders );
```
schlessera: I'm probably missing something here, but I think it warrants a comment. The goal here is to verify the ordering of the placeholders. However, before comparing them, you sort them here. Would that put them into the same order no matter what?

swissspidy: I don't know what I was thinking here. I copied this from WPCS, but somehow messed it up. The order should be irrelevant, but different placeholders are usually not ok.

swissspidy:
I think this one is good to go (pending passing tests and final review).

Nice side effect when this is in master:

I can use #69 to create a WordPress Trac ticket to improve all the strings in WordPress core that get flagged by this command.

Issue #55, opened 2018-07-25
Add description comment to POT file


swissspidy:
As noted in https://github.com/wp-cli/i18n-command/issues/47#issuecomment-407429435:

I just noticed that the original makepot.php script has a [description field](https://github.com/WordPress/wordpress-develop/blob/335b69cedc8a828af1e2c4fa841fcafac6700fbb/tools/i18n/makepot.php#L85-L144) for projects, e.g. `Translation of the WordPress theme {name} {version} by {author}`. However, that appears to be unused.

We might want to add this as a default comment to the POT file though as it can be useful when opening a POT file without context.

Issue #47, opened 2018-07-17
Allow passing custom headers and comments for POT file


greatislander:
Currently, `wp i18n make-pot` generates a `.pot` file and adds a link to the WordPress plugin or theme repo page as the `Report-Msgid-Bugs-To` header.

See:

+ https://github.com/wp-cli/i18n-command/blob/master/src/MakePotCommand.php#L375
+ https://github.com/wp-cli/i18n-command/blob/master/src/MakePotCommand.php#L379

For plugins and themes which are not distributed via WordPress repos, adding an optional parameter such as `--bugs_address` could be used to supply an alternate link (e.g. to a GitHub issues page).

swissspidy:
Tools like [grunt-wp-i18n](https://github.com/cedaro/grunt-wp-i18n/blob/develop/docs/makepot.md) also allow specifying a custom copyright comment notice. Would that be needed too in your case?

greatislander:
@swissspidy That would be useful as well, yes.

swissspidy:
Related: https://github.com/wp-cli/wp-cli/issues/4616, https://github.com/wp-cli/entity-command/pull/133

We could add a `--headers` argument that allows passing an associative array in JSON format to add custom headers.

Example:

```
wp i18n make-pot --headers='{"Report-Msgid-Bugs-To":"https://github.com/wp-cli/i18n-command/issues"}'
```

This could also be useful to add other headers if one really wants to (see #48 and #49).

For the copyright notice I'd add a new `--copyright` argument (since it's a comment and not a header) or perhaps a new `--comments` argument that takes a comma separated list of comments to add (overriding the existing copyright notice if used). This would make things much more flexible.

----

I just noticed that the original makepot.php script has a [description field](https://github.com/WordPress/wordpress-develop/blob/335b69cedc8a828af1e2c4fa841fcafac6700fbb/tools/i18n/makepot.php#L85-L144) for projects, e.g. `Translation of the WordPress theme {name} {version} by {author}`. However, that appears to be unused.

We might want to add this as a default comment to the POT file though as it can be useful when opening a POT file without context.

greatislander:
I like the JSON `--headers` idea. Much more flexible than adding a bunch of parameters. I'd be happy to work on a PR for that. And I'd lean towards the `--comments` parameter (again, over specificity that could be limiting).

PR #69, opened 2018-08-05
WordPress Core support

### LINKED ISSUE #36: Explore usefulness for WordPress core

### LINKED ISSUE #45: Ignore minified JavaScript files
Right now, we exclude some files by default. One file pattern that we don't yet exclude is minified JavaScript files. They usually look like `i-am-minified.min.js`.

It's not currently possible to exclude files using glob patterns like `*.min.js` but I think that would be a great addition.

When someone excludes `*.min.js`, we can turn this into a regex à la `.*\.min.js` or something, and then exclude the file if it matches that regex. `WP_oEmbed::get_provider()` has some code that does this. Perhaps we can reuse this.

In the PHP extractor, the relevant code is here: https://github.com/wp-cli/i18n-command/blob/bd16b5fd631980dea4684e9b290b3e80674b4401/src/PhpCodeExtractor.php#L129-L146

We can probably extend this with a new regex.

swissspidy:
Changes in this PR:

* Allows passing more than one file to `--merge`
* Does not error when no plugin or theme file has been found
* Adds glob support to `--exclude` (see #45)
* Adds a new `--include` argument as an addition to `--exclude`
* Adds new `--copyright-holder` and `--package-name` arguments  
  These values are used for the copyright notice comment at the top of the POT file.
  The naming of the arguments and the usefulness of that comment as a whole should be discussed.
* Adds a new `--except` option that accepts one or more POT files  
  Strings that are on this "blacklist" won't be extracted

----

**To do:**

* [x] Add necessary features for supporting string extraction in WordPress core (and perhaps any other project)
* [x] Cover these new features in tests

----

The following commands can be used to generate the necessary POT files for WordPress core.

Note: the `Report-Msgid-Bugs-To` URL can be overridden using the `--header` argument.

**Continents & Cities**

```bash
wp i18n make-pot /path/to/wordpress/core /path/to/target/wp-tz.pot \
--include="wp-admin/includes/continents-cities.php" \
--package-name="WordPress" \
--copyright-holder="WordPress" \
--debug \
--skip-js \
--ignore-domain
```

**WordPress** (Development 4.9.x on translate.wordpress.org)

```bash
wp i18n make-pot /path/to/wordpress/core /path/to/target/wordpress.pot \
--exclude="wp-admin/*,wp-content/themes/*,wp-includes/class-pop3.php,wp-content/plugins/akismet/" \
--package-name="WordPress" \
--copyright-holder="WordPress" \
--debug \
--skip-js \
--ignore-domain
```

**Administration**

This one involves some more steps as it should include all files except for _WordPress_ and _Network Admin_  files. Also, the Hello Dolly plugin, which is bundled with WordPress core, needs to be translated as well.

1. Hello Dolly:

```bash
wp i18n make-pot /path/to/wordpress/core/wp-content/plugins /path/to/target/hello-dolly.pot \
--include="hello.php" \
--exclude="akismet" \
--debug \
--skip-js \
--ignore-domain
```

2. WordPress Administration

```bash
wp i18n make-pot /path/to/wordpress/core /path/to/target/wp-admin.pot \
--exclude="wp-admin/network/*,wp-admin/network.php,wp-admin/includes/class-wp-ms*,wp-admin/includes/network.php,wp-admin/includes/continents-cities.php" \
--include="wp-admin/*" \
--merge="/path/to/target/hello-dolly.pot" \
--except="/path/to/target/wp-frontend.pot" \
--package-name="WordPress" \
--copyright-holder="WordPress" \
--debug \
--skip-js \
--ignore-domain
```

**Network Admin**

```bash
wp i18n make-pot /path/to/wordpress/core /path/to/target/wp-network-admin.pot \
--include="wp-admin/network/*,wp-admin/network.php,wp-admin/includes/class-wp-ms*,wp-admin/includes/network.php" \
--except="/path/to/target/wp-frontend.pot,/path/to/target/wp-admin.pot" \
--package-name="WordPress" \
--copyright-holder="WordPress" \
--debug \
--skip-js \
--ignore-domain
```

----

Fixes #36.
Fixes #45.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ wp i18n
 Create a POT file for a WordPress plugin or theme.
 
 ~~~
-wp i18n make-pot <source> [<destination>] [--slug=<slug>] [--domain=<domain>] [--ignore-domain] [--merge[=<file>]] [--exclude=<paths>] [--skip-js]
+wp i18n make-pot <source> [<destination>] [--slug=<slug>] [--domain=<domain>] [--ignore-domain] [--merge[=<files>]] [--except=<files>] [--include=<paths>] [--exclude=<paths>] [--headers=<headers>] [--skip-js] [--copyright-holder=<name>] [--package-name=<name>]
 ~~~
 
 Scans PHP and JavaScript files, as well as theme stylesheets for translatable strings.
@@ -55,9 +55,16 @@ Scans PHP and JavaScript files, as well as theme stylesheets for translatable st
 	[--ignore-domain]
 		Ignore the text domain completely and extract strings with any text domain.
 
-	[--merge[=<file>]]
-		Existing POT file file whose content should be merged with the extracted strings.
+	[--merge[=<files>]]
+		One or more existing POT files whose contents should be merged with the extracted strings.
 		If left empty, defaults to the destination POT file.
+
+	[--except=<files>]
+		If set, only strings not already existing in one of the passed POT files will be extracted.
+
+	[--include=<paths>]
+		Only take specific files and folders into account for the string extraction.
+		Leading and trailing slashes are ignored, i.e. `/my/directory/` is the same as `my/directory`.
 
 	[--exclude=<paths>]
 		Include additional ignored paths as CSV (e.g. 'tests,bin,.github').
@@ -69,6 +76,12 @@ Scans PHP and JavaScript files, as well as theme stylesheets for translatable st
 
 	[--skip-js]
 		Skips JavaScript string extraction. Useful when this is done in another build step, e.g. through Babel.
+
+	[--copyright-holder=<name>]
+		Name to use for the copyright comment in the resulting POT file.
+
+	[--package-name=<name>]
+		Name to use for package name in the resulting POT file. Overrides anything found in a plugin or theme.
 
 **EXAMPLES**
 
--- a/README.md
+++ b/README.md
@@ -28,13 +28,14 @@ wp i18n
 
 ### wp i18n make-pot
 
-Create a POT file for a WordPress plugin or theme.
+Create a POT file for a WordPress project.
 
 ~~~
-wp i18n make-pot <source> [<destination>] [--slug=<slug>] [--domain=<domain>] [--ignore-domain] [--merge[=<files>]] [--except=<files>] [--include=<paths>] [--exclude=<paths>] [--headers=<headers>] [--skip-js] [--copyright-holder=<name>] [--package-name=<name>]
+wp i18n make-pot <source> [<destination>] [--slug=<slug>] [--domain=<domain>] [--ignore-domain] [--merge[=<paths>]] [--subtract=<paths>] [--include=<paths>] [--exclude=<paths>] [--headers=<headers>] [--skip-js] [--file-comment] [--package-name=<name>]
 ~~~
 
-Scans PHP and JavaScript files, as well as theme stylesheets for translatable strings.
+Scans PHP and JavaScript files for translatable strings, as well as theme stylesheets and plugin files
+if the source directory is detected as either a plugin or theme.
 
 **OPTIONS**
 
@@ -50,26 +51,34 @@ Scans PHP and JavaScript files, as well as theme stylesheets for translatable st
 	[--domain=<domain>]
 		Text domain to look for in the source code, unless the `--ignore-domain` option is used.
 		By default, the "Text Domain" header of the plugin or theme is used.
-		If none is provided, it falls back to the plugin/theme slug.
+		If none is provided, it falls back to the project slug.
 
 	[--ignore-domain]
 		Ignore the text domain completely and extract strings with any text domain.
 
-	[--merge[=<files>]]
-		One or more existing POT files whose contents should be merged with the extracted strings.
-		If left empty, defaults to the destination POT file.
+	[--merge[=<paths>]]
+		Comma-separated list of POT files whose contents should be merged with the extracted strings.
+		If left empty, defaults to the destination POT file. POT file headers will be ignored.
 
-	[--except=<files>]
-		If set, only strings not already existing in one of the passed POT files will be extracted.
+	[--subtract=<paths>]
+		Comma-separated list of POT files whose contents should act as some sort of blacklist for string extraction.
+		Any string which is found on that blacklist will not be extracted.
+		This can be useful when you want to create multiple POT files from the same source directory with slightly
+		different content and no duplicate strings between them.
 
 	[--include=<paths>]
-		Only take specific files and folders into account for the string extraction.
+		Comma-separated list of files and paths that should be used for string extraction.
+		If provided, only these files and folders will be taken into account for string extraction.
+		For example, `--include="src,my-file.php` will ignore anything besides `my-file.php` and files in the `src` directory.
+		Simple glob patterns can be used, i.e. `--include=foo-*.php` includes any PHP file with the `foo-` prefix.
 		Leading and trailing slashes are ignored, i.e. `/my/directory/` is the same as `my/directory`.
 
 	[--exclude=<paths>]
-		Include additional ignored paths as CSV (e.g. 'tests,bin,.github').
-		By default, the following files and folders are ignored: node_modules, .git, .svn, .CVS, .hg, vendor.
+		Comma-separated list of files and paths that should be skipped for string extraction.
+		For example, `--exclude=".github,myfile.php` would ignore any strings found within `myfile.php` or the `.github` folder.
+		Simple glob patterns can be used, i.e. `--exclude=foo-*.php` excludes any PHP file with the `foo-` prefix.
 		Leading and trailing slashes are ignored, i.e. `/my/directory/` is the same as `my/directory`.
+		The following files and folders are always excluded: node_modules, .git, .svn, .CVS, .hg, vendor, *.min.js.
 
 	[--headers=<headers>]
 		Array in JSON format of custom headers which will be added to the POT file. Defaults to empty array.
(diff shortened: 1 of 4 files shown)


swissspidy:
Here's the output of the WP-CLI command and `makepot.php` for a quick comparison:

[makepot.zip](https://github.com/wp-cli/i18n-command/files/2260556/makepot.zip)
[wp-cli-i18n-command.zip](https://github.com/wp-cli/i18n-command/files/2260557/wp-cli-i18n-command.zip)

The POT files for continents and cities have the same number of strings (524). The others seem to differ quite a bit.

Tested both with current WordPress master/trunk

| Type                | WP-CLI | makepot.php |
|---------------------|--------|-------------|
| Frontend            | 2917   | 2665        |
| Continents & Cities | 524    | 524         |
| Administration      | 2975   | 2645        |
| Network Admin       | 405    | 311         |

I took the numbers on the right from [translate.wordpress.org](https://translate.wordpress.org/locale/de/default/wp/dev) as the local output with `makepot.php` seems to be quite different. Not sure why though.

Perhaps @ocean90 can help me find the cause for this :-)

swissspidy:
Instead of the nightly build I ran this now against 4.9.8 as downloaded straight from [WordPress.org/download](https://wordpress.org/download/).

The numbers are much better now:

| Type                | WP-CLI | makepot.php |
|---------------------|--------|-------------|
| Frontend            | 2665   | 2665        |
| Continents & Cities | 524    | 524         |
| Administration      | 2981   | 2645        |
| Network Admin       | 404    | 311         |

Administration now includes Hello Dolly (as does `makepot.php` / translate.wordpress.org), that's why that number is higher than before.

swissspidy:
So, the reason for the difference between the admin files seems to be that `makepot.php` removes any strings that are already part of the frontend POT file to avoid any duplicates.

It uses the [`msgcat`](https://www.gnu.org/software/gettext/manual/html_node/msgcat-Invocation.html) utility for that. So when generating the admin POT file it actually does this:

1. Generate frontend POT file
2. Generate admin POT file
3. Merge frontend with admin file and only keep the common strings
4. Merge admin file with the common file and only keep the unique strings

I'm not sure this is something that could or should be done within this command.

I think if WordPress.org is going to use the WP-CLI command at some point for string extraction, they might add a little wrapper script around it to use `msgcat` as outlined above to keep existing behavior.

schlessera:
Instead of using `msgcat`, maybe it would be worthwhile to add a `--deduplicate` flag to `--merge` ?

swissspidy:
@schlessera So you'd pass a file name to `--deduplicate`, e.g. `--deduplicate=wp-frontend.pot`?

Then we could use something like this:

```bash
wp i18n make-pot /path/to/wordpress/core /path/to/target/wp-admin.pot \
--exclude="wp-admin/network/*,wp-admin/network.php,wp-admin/includes/class-wp-ms*,wp-admin/includes/network.php,wp-admin/includes/continents-cities.php" \
--include="wp-admin/*" \
--merge="/path/to/target/hello-dolly.pot" \
--deduplicate="/path/to/target/wp-frontend.pot" \
--package-name="WordPress" \
--copyright-holder="WordPress" \
--debug \
--skip-js \
--ignore-domain

**Comment by schlessera** [comment]
I'm not sure, I ignore how the files relate to each other in detail, and what will be considered the source of truth.

But generally, if you have a `--merge` command, it makes sense in my hand to specify how duplicates are handled.

**Comment by schlessera** [comment]
Hmm, I see the problem now. What you need is not deduplication, but set operations like union and most of all intersection.

**Comment by swissspidy** [comment]
I added an experimental `--except` argument now.

You can pass one or more exceptions. The command extracts all strings from them and compares them against the ones found by the command.

Strings that are contained in the exception list will be removed from final output.

This way, I got the resulting POT file for wp-admin down to 2733 strings (compared to 2981 before).

**Comment by swissspidy** [comment]
Alright, got it down to 2647! That's only 2 differences.

----

Note regarding exceptions: this might be useful as a new merge flag in the upstream gettext library instead, see https://github.com/oscarotero/Gettext#merge-translations.

**Comment by swissspidy** [comment]
I'm not sure why, but the output generated by `makepot.php` seems to be missing the following two strings:

* `<strong>ERROR</strong>: "Table Prefix" can only contain numbers, letters, and underscores.`  
  source: `wp-admin/setup-config.php:262`
* `<strong>ERROR</strong>: "Table Prefix" must not be empty.`  
  source: `wp-admin/setup-config.php:258`

I can't find these strings on https://translate.wordpress.org either.

Sounds like a bug to me and a reason why this command here should be used instead :-)

**Comment by swissspidy** [comment]
**Updated and final numbers:**

| Type                | WP-CLI | makepot.php |
|---------------------|--------|-------------|
| Frontend            | 2665   | 2665        |
| Continents & Cities | 524    | 524         |
| Administration      | 2647   | 2645        |
| Network Admin       | 311    | 311         |

----

I feel like now we only need to discuss the wording of the newly added arguments or, in the case of the copyright notice, the usefulness of these.

**Comment by schlessera** [comment]
Regarding the `--except` flag, I think the name is slightly misleading. I feel like `--subtract` might be clearer, wouldn't it?

File: README.md [hunk 1]
```diff
@@ -55,10 +55,17 @@ Scans PHP and JavaScript files, as well as theme stylesheets for translatable st
 	[--ignore-domain]
 		Ignore the text domain completely and extract strings with any text domain.
 
-	[--merge[=<file>]]
-		Existing POT file file whose content should be merged with the extracted strings.
+	[--merge[=<files>]]
+		One or more existing POT files whose contents should be merged with the extracted strings.
```
schlessera: The syntax for multiple files should be documented.

README.md
```diff
@@ -55,10 +55,17 @@ Scans PHP and JavaScript files, as well as theme stylesheets for translatable st
...
 		If left empty, defaults to the destination POT file.
 
+	[--except=<files>]
```
schlessera: As already commented, I think `--subtract` would be a clearer name.
Also, syntax for multiples files should be documented.

README.md
```diff
@@ -55,10 +55,17 @@ Scans PHP and JavaScript files, as well as theme stylesheets for translatable st
...
+		If set, only strings not already existing in one of the passed POT files will be extracted.
+
+	[--include=<paths>]
```
schlessera: Syntax for multiple paths needs to be documented.

features/makepot.feature
```diff
@@ -1530,3 +1718,149 @@ Feature: Generate a POT file of a WordPress project
...
+      msgid "Some other text"
+      """
+
+  Scenario: Extract strings for a generic project
+    Given an empty example-project directory
+    And a example-project/stuff.php file:
+      """
+      <?php
+
+       __( 'Hello World' );
```
schlessera: A regular project would normally use regular gettext with single underscore: `_()`

src/MakePotCommand.php
```diff
@@ -97,51 +124,71 @@ class MakePotCommand extends WP_CLI_Command {
...
-		$this->slug      = Utils\get_flag_value( $assoc_args, 'slug', Utils\basename( $this->source ) );
-		$this->skip_js   = Utils\get_flag_value( $assoc_args, 'skip-js', $this->skip_js );
-		$this->headers   = Utils\get_flag_value( $assoc_args, 'headers', $this->headers );
+		$assoc_args      = Utils\parse_shell_arrays( $assoc_args, $array_arguments );
+
+		$this->source           = realpath( $args[0] );
+		$this->slug             = Utils\get_flag_value( $assoc_args, 'slug', Utils\basename( $this->source ) );
+		$this->skip_js          = Utils\get_flag_value( $assoc_args, 'skip-js', $this->skip_js );
+		$this->headers          = Utils\get_flag_value( $assoc_args, 'headers', $this->headers );
+		$this->package_name     = Utils\get_flag_value( $assoc_args, 'package-name', 'Unknown' );
```
schlessera: Shouldn't the package and copyright holder be optional?
If I understand correctly, this will now always be added, but with values of `'Unknown'`.

swissspidy:
It‘s certainly not ideal, yes. I have yet to come up with a better name. Subtract could work.

README.md
swissspidy: Do you mean documented like `<file1,file2>` or as an example in the argument description?

src/MakePotCommand.php
swissspidy: We could either make it optional, change the wording (just „Copyright 2018“ without any author or name) or remove it completely. I‘m not sure if it really serves any purpose.

Also, two command arguments just for this comment is a bit too much I think. Perhaps we could go with a singular argument to manually override that line?

README.md
schlessera: Not sure what the best way is, as you have multiple arguments with the same issue.

Maybe a small adjective like `comma-separated` and then a general note in the command description might be clearer?

```
	[--merge[=<file[,file2,...]>]]
		Comma-separated list of POT files whose contents should be merged with the extracted strings.
```

Something like that, maybe? Unsure about the notation...

src/MakePotCommand.php
schlessera: Yes, something more generic like `--file-comment` might be preferable.

features/makepot.feature
swissspidy: I can add both, although I don‘t think we should enphasise support for any regular, non-WP gettext project.

src/MakePotCommand.php
swissspidy: See https://github.com/wp-cli/i18n-command/issues/47#issuecomment-407429435 and #55 for previous discussion about a `--comment` / `--file-comment` argument

schlessera: I think `--comment` could be ambiguous, as it could just as well refer to a translator comment.

features/makepot.feature
schlessera: It was more about having an example that is technically correct. Outside of WordPress, `__()` would normally not exist.

README.md
swissspidy: Turns out something like `[--merge[=<file[,file2,...]>]]` isn't valid.

I'll stick with the description though.

src/MakePotCommand.php
swissspidy: I added `--file-comment` but left `package-name` as it's needed/useful for the `Project-Id-Version` POT header.

schlessera:
:tada:

johnbillion:
@swissspidy Regarding the two "Table prefix" strings that makepot missed, it might be because they were malformed. See https://core.trac.wordpress.org/changeset/43205.

PR #77, opened 2018-08-12
Automatically rerun failed scenarios

wojsmol:
fixes https://github.com/wp-cli/i18n-command/issues/75


composer.json
```diff
@@ -45,6 +45,7 @@
     "prefer-stable": true,
     "scripts": {
         "behat": "run-behat-tests",
+		"behat-rerun": "rerun-behat-tests",
```
swissspidy: Wrong indentation

PR #82, opened 2018-09-13
Allow passing an empty file comment

swissspidy:
Related: https://github.com/Automattic/amp-wp/issues/1416.

This change allows omitting the file comment at the top of the POT file to prevent issues with the pot-to-php script that Gutenberg and other plugins use.

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -2011,3 +2011,33 @@ Feature: Generate a POT file of a WordPress project
       # Copyright (C) 2018 John Doe
       # Powered by WP-CLI.
       """
+
+  Scenario: Empty file comment
+    Given an empty example-project directory
+    And a example-project/stuff.php file:
+      """
+      <?php
+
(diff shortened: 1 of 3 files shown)


features/makepot.feature
```diff
@@ -2011,3 +2011,33 @@ Feature: Generate a POT file of a WordPress project
...
+       * License:     GPL-2.0+
+       * License URI: http://www.gnu.org/licenses/gpl-2.0.txt
+       * Text Domain: foo-plugin
+       * Domain Path: /languages
+       */
+
+       __( 'Hello World', 'foo-plugin' );
+      """
+
+    When I run `wp i18n make-pot example-project result.pot --ignore-domain --file-comment="d"`
```
schlessera: What is the `"d"` file comment here? Looks like a typo... I suppose the scenario is meant to test `""`, not `"d"`.

Issue #84, opened 2018-09-21
Not possible to include `vendor`


remcotolsma:
It looks like it's not possible to include the `vendor` folder in the `make-pot` command:

- https://github.com/wp-cli/i18n-command/blob/v2.0.1/src/MakePotCommand.php#L41-L44
- https://github.com/wp-cli/i18n-command/blob/v2.0.1/src/MakePotCommand.php#L375-L379
- https://github.com/wp-cli/i18n-command/blob/v2.0.1/src/MakePotCommand.php#L527-L533
- https://github.com/wp-cli/i18n-command/blob/v2.0.1/src/IterableCodeExtractor.php#L173-L179

We have a few plugins where we require to include the `vendor` folder, how can we do this?

swissspidy:
You're right, it's currently not possible to override built-in excludes.

However, it was a conscious decision to only exclude a specific, limited set of files and folders.

What is your use case exactly?

`vendor` is typically used by Composer. In your theme/plugin you can add third-party libraries as dependencies and they will end up in that folder. These dependencies should obviously have different text domains than your project and you shouldn't have to care about translations of a dependency anyway.

remcotolsma:
We have split a WordPress plugin into a number of Composer packages, but they share the same text domain. Probably not the best design, but for now we can't change this very easily. I also don't know how it works with multiple text domains in one plugin and the https://translate.wordpress.org/projects/wp-plugins/ translation platform. For now fixed it with a custom `make-pot` command:

```
<?php

namespace Pronamic\WordPress\Pay;

class MakePotCommand extends \WP_CLI\I18n\MakePotCommand {
	public function __construct() {
		parent::__construct();

		// https://github.com/wp-cli/i18n-command/blob/v2.0.1/src/MakePotCommand.php#L36-L44
		$this->exclude = array_diff( $this->exclude, array(
			'vendor',
		) );

		$this->exclude = array_merge( $this->exclude, array(
			'build',
			'deploy',
			'documentation',
			'etc',
			'repositories',
			'wordpress',
			'wp-content',
		) );

		$this->include = array(
			'admin',
			'includes',
			'templates',
			'vendor',
			'views',
		);
	}
}

// https://github.com/wp-cli/i18n-command/blob/v2.0.1/i18n-command.php
\WP_CLI::add_command( 'pronamic i18n make-pot', '\Pronamic\WordPress\Pay\MakePotCommand' );
[...]

Issue #70, opened 2018-08-06
Use unit tests for  `--include` and `--exclude` syntax instead of functional tests


schlessera:
Some of tests, including but maybe not limited to `--include` and `--exclude` syntax tests, would be a better fit as unit tests. These would be more reliable and much faster.

swissspidy:
There are now some unit tests since #83.

We might clean up the folder structure for them at some point though.

Also, there are still some Behat tests for these arguments, but I think we can keep them for now as it doesn't matter much.

PR #85, opened 2018-10-24
Support parsing minified JS files created by WebPack

herregroen:
In order to resolve https://meta.trac.wordpress.org/ticket/3875 `wp i18n make-pot` should be able to traverse minified files created by WebPack in order to find where each translation is used, this so that when generating language packs it can generate a file for each JS file containing only the necessary translations.

Because WebPack wraps calls to `wp.i18n.__( "string" )` in the following form `Object(u.__)( "string" )` this adds support for `CallExpression`s where the callee is a `CallExpression` itself whose callee is `Object` and argument the necessary translation function.

This part will likely later need to be expanded to support more forms in which translations may be called. ~~Supporting `MemberExpression` as the callee for example would be an obvious first improvement.~~ Support for `MemberExpression` as the callee has been added in the PR.

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -1431,11 +1431,13 @@ Feature: Generate a POT file of a WordPress project
       );
 
       wp.i18n.__( 'wp.i18n.__', 'foo-plugin' );
+      wp.i18n._n( 'wp.i18n._n_single', 'wp.i18n._n_plural', number, 'foo-plugin' );
 
       const translate = wp.i18n;
       translate.__( 'translate.__', 'foo-plugin' );
 
       Object(u.__)( 'minified.__', 'foo-plugin' );
+      Object(j._x)( 'minified._x', 'minified._x_context', 'foo-plugin' );
       """
 
     When I run `wp i18n make-pot foo-plugin`
@@ -1487,11 +1489,27 @@ Feature: Generate a POT file of a WordPress project
       """
     And the foo-plugin/foo-plugin.pot file should contain:
       """
+      msgid "wp.i18n._n_single"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgid_plural "wp.i18n._n_plural"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
       msgid "translate.__"
       """
     And the foo-plugin/foo-plugin.pot file should contain:
       """
       msgid "minified.__"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgid "minified._x"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgctxt "minified._x_context"
       """
     And the foo-plugin/foo-plugin.pot file should not contain:
       """
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -1429,6 +1429,13 @@ Feature: Generate a POT file of a WordPress project
         <h1>{__( 'Hello JSX', 'foo-plugin' )}</h1>,
         document.getElementById('root')
       );
+
+      wp.i18n.__( 'wp.i18n.__', 'foo-plugin' );
+
+      const translate = wp.i18n;
+      translate.__( 'translate.__', 'foo-plugin' );
+
+      Object(u.__)( 'minified.__', 'foo-plugin' );
       """
 
     When I run `wp i18n make-pot foo-plugin`
@@ -1473,6 +1480,18 @@ Feature: Generate a POT file of a WordPress project
     And the foo-plugin/foo-plugin.pot file should contain:
       """
       msgid "Hello JSX"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgid "wp.i18n.__"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgid "translate.__"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgid "minified.__"
       """
     And the foo-plugin/foo-plugin.pot file should not contain:
       """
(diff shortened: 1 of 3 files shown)


herregroen:
@schlessera Thanks for the review!

I've added tests to makepot.feature. These are currently simply added to existing JS parsing test as especially following `MemberExpression` callees ( for example: `wp.i18n.__` ) isn't exclusive to minified files.

The WebPack specific format ( `Object(u.__)( 'translation' )` is a lot more niche so let me know if you feel there should be a separate scenario for that.

Resolving the callee has also been split to it's own function with more extensive documentation and examples.

src/JsFunctionsScanner.php
```diff
@@ -152,4 +151,44 @@ public function saveGettextFunctions( Translations $translations, array $options
...
+			return $node->getCallee()->getProperty();
+		}
+
+		// If the callee is a call expression as created by WebPack resolve it.
+		// For example: Object(u.__)( "translation" ).
+		if (
+			'CallExpression' ===  $node->getCallee()->getType() &&
+			'Identifier' ===  $node->getCallee()->getCallee()->getType() &&
+			'Object' === $node->getCallee()->getCallee()->getName() &&
+			! empty( $node->getCallee()->getArguments() ) &&
```
schlessera: This fails on PHP 5.4: https://travis-ci.org/wp-cli/i18n-command/jobs/446078984#L633

`empty()` needs to work on a reference for PHP 5.4 and lower, so you cannot directly use a method call here.

swissspidy:
Instead of parsing `*.min.js` by default, I think we should leave the `exclude` rule as-is, but make it possible to `include` them.

For this we'd need to swap the include and exclude check as discussed in #84. I'm gonna try to merge the unit tests for that (#70) and then attempt modifying `IterableCodeExtractor::getFilesFromDirectory()` accordingly.

src/JsFunctionsScanner.php
schlessera: Alternatively, you can change to test to not use `empty()`, but compare to an empty array (which I think is the expected return value here).

src/MakePotCommand.php
```diff
@@ -41,7 +41,7 @@ class MakePotCommand extends WP_CLI_Command {
 	/**
 	 * @var array
 	 */
-	protected $exclude = [ 'node_modules', '.git', '.svn', '.CVS', '.hg', 'vendor', 'Gruntfile.js', 'webpack.config.js', '*.min.js' ];
+	protected $exclude = [ 'node_modules', '.git', '.svn', '.CVS', '.hg', 'vendor', 'Gruntfile.js', 'webpack.config.js' ];
```
schlessera: As #87 was now merged, this change is not needed anymore.

By default, minified files should still be excluded, and if you want to explicitly include them, you can do so by adding `--include=*.min.js` to the command.

src/JsFunctionsScanner.php
```diff
@@ -152,4 +151,44 @@ public function saveGettextFunctions( Translations $translations, array $options
...
+		// If the callee is a member expression resolve it to the property.
+		// For example: wp.i18n.__( "translation" ) or u.__( "translation" ).
+		if (
+			'MemberExpression' === $node->getCallee()->getType() &&
+			'Identifier' === $node->getCallee()->getProperty()->getType()
+		) {
+			return $node->getCallee()->getProperty();
+		}
+
+		// If the callee is a call expression as created by WebPack resolve it.
```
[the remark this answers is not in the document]
schlessera: You can do spiffy suggestions now, which is usually quicker for both parties involved in cases like this:
```suggestion
		// If the callee is a call expression as created by Webpack resolve it.
```
@herregroen will then be able to just accept this with one click.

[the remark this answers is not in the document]
swissspidy: @schlessera The last few times I tried it, the suggestion feature completely failed and it committed things like `span` tags in addition to the actual fix. That's why I currently hesitate to use it 🙂

features/makepot.feature
```diff
@@ -1429,6 +1429,13 @@ Feature: Generate a POT file of a WordPress project
         <h1>{__( 'Hello JSX', 'foo-plugin' )}</h1>,
         document.getElementById('root')
       );
+
+      wp.i18n.__( 'wp.i18n.__', 'foo-plugin' );
+
+      const translate = wp.i18n;
+      translate.__( 'translate.__', 'foo-plugin' );
+
+      Object(u.__)( 'minified.__', 'foo-plugin' );
```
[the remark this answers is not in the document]
herregroen: Yes, all variations work. I'll expand the test suite to cover those.

src/MakePotCommand.php
herregroen: Awesome! Will remove this in that case.

src/JsFunctionsScanner.php
```diff
@@ -152,4 +151,46 @@ public function saveGettextFunctions( Translations $translations, array $options
...
+			return $callee->getProperty();
+		}
+
+		// If the callee is a call expression as created by Webpack resolve it.
+		// For example: Object(u.__)( "translation" ).
+		if (
+			'CallExpression' ===  $callee->getType() &&
+			'Identifier' ===  $callee->getCallee()->getType() &&
+			'Object' === $callee->getCallee()->getName() &&
+			array() !== $callee->getArguments() &&
```
swissspidy:
```suggestion
			[] !== $callee->getArguments() &&
```

ocean90:
Using this PR and I can't get the JS strings of the current 5.0 branch  into the POT file. I'm using this command:

```
php -d memory_limit=512M "$(which wp)" i18n make-pot ~/wordpress-50 ~/wordpress-50.pot --exclude="wp-admin/*,wp-content/themes/*,wp-includes/class-pop3.php,wp-content/plugins/akismet/" --package-name='WordPress' --headers='{"Report-Msgid-Bugs-To":"https://core.trac.wordpress.org/","PO-Revision-Date":"YEAR-MO-DA HO:MI+ZONE"}' --file-comment='Copyright (C) 2018 by the contributors\nThis file is distributed under the same license as the WordPress package.' --ignore-domain
```

Am I missing something?

schlessera:
@ocean90 If the files are minified, you need to include them explicitly with `--include=*.min.js`.

ocean90:
This is now working for me, after updating to the latest revisions and using the correct packages directory....

PR #87, opened 2018-10-25
Include should override exclude instead of the other way around

### LINKED ISSUE #84: Not possible to include `vendor`

swissspidy:
This is still a WIP as some Behat tests are failing.

Any help appreciated as I might be able to work on it again until Saturday.

Fixes #84.
See also #85.

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -1161,8 +1161,14 @@ Feature: Generate a POT file of a WordPress project
       __( 'I am not being ignored', 'foo-plugin' );
       """
 
-    When I run `wp i18n make-pot foo-plugin foo-plugin.pot --include=*.min.js`
+    When I run `wp i18n make-pot foo-plugin foo-plugin.pot`
     Then the foo-plugin.pot file should not contain:
+      """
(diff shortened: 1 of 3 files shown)


schlessera:
The issue seems to be that you're using absolute paths in some places, instead of root-relative paths.

The `vendor` folder therefore gets pulled in (instead of being excluded because the include `foo` matches `<absolute path>/foo-plugin/vendor`.

I'll look at adapting the code.

PR #99, opened 2018-10-29
Replace regexes Peast can't parse.

### LINKED ISSUE #98: Regular expression literal in JS code can be mis-parsed as unterminated strings
The JavaScript parsing code contains a bug in the upstream [mck89/peast](https://github.com/mck89/peast) library.

If the parsed code has regular expression literals containing quotes, these quotes are erroneously being parsed as opening or closing a string literal.

This will in turn cause the parser to throw an `Unterminated string` error once it hits the end of the line.

As an example, the WordPress Core code contains regular expression literals like this:

```js
function escapeHTML(s) {
   var n = s;
   n = n.replace(/&/g, '&amp;');
   n = n.replace(/</g, '&lt;');
   n = n.replace(/>/g, '&gt;');
   n = n.replace(/"/g, '&quot;');

   return n;
}
```

In this example, the regular expression literal `/"/g` will be parsed as opening a double-quoted string and throw an error once it hits the next line.

herregroen:
Fixes #98.

Currently only covers the simple `/"/` use-case which is the only one we're encountering right now.

Could be updated to use a regex to cover more complex scenarios but I'd prefer to avoid that. Should ideally be fixed in https://github.com/mck89/peast/issues/3.

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -1474,7 +1474,7 @@ Feature: Generate a POT file of a WordPress project
       """
     And a foo-plugin/foo-plugin.js file:
       """
-      // Includes to test if peast correctly parses regexes containing a quote.
+      // Included to test if peast correctly parses regexes containing a quote.
       // See: https://github.com/wp-cli/i18n-command/issues/98
       n = n.replace(/"/g, '&quot;');
 
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -1474,6 +1474,10 @@ Feature: Generate a POT file of a WordPress project
       """
     And a foo-plugin/foo-plugin.js file:
       """
+      // Includes to test if peast correctly parses regexes containing a quote.
+      // See: https://github.com/wp-cli/i18n-command/issues/98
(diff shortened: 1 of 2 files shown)


features/makepot.feature
```diff
@@ -1474,6 +1474,10 @@ Feature: Generate a POT file of a WordPress project
       """
     And a foo-plugin/foo-plugin.js file:
       """
+      // Includes to test if peast correctly parses regexes containing a quote.
```
schlessera:
```suggestion
      // Included to test if peast correctly parses regexes containing a quote.
```

src/JsFunctionsScanner.php
```diff
@@ -37,6 +37,9 @@ public function disableCommentsExtraction() {
 	 * {@inheritdoc}
 	 */
 	public function saveGettextFunctions( Translations $translations, array $options ) {
+		// Replace unparseable regexes
```
schlessera:
```suggestion
		// Temporary workaround for https://github.com/wp-cli/i18n-command/issues/98.
```

PR #103, opened 2018-10-30
Add parsing of .map files to improve reliability of JS translations

herregroen:
Adds a new `MapCodeExtractor` to parse `.map` files ( that may accompany minified JS files ).

It will parse these using the existing `JsFunctionsScanner` to improve picking up translations and translator's comments that may get lost during build steps and/or minification.

### DIFF
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -554,6 +554,16 @@ class MakePotCommand extends WP_CLI_Command {
 						'extensions' => [ 'js' ],
 					]
 				);
+
+				MapCodeExtractor::fromDirectory(
+					$this->source,
+					$translations,
+					[
+						'include'    => $this->include,
+						'exclude'    => $this->exclude,
+						'extensions' => [ 'map' ],
+					]
+				);
 			}
 		} catch ( \Exception $e ) {
 			WP_CLI::error( $e->getMessage() );
--- a/src/MapCodeExtractor.php
+++ b/src/MapCodeExtractor.php
@@ -48,7 +48,7 @@ final class MapCodeExtractor extends JsCode {
 		} catch ( PeastException $e ) {
 			WP_CLI::debug(
 				sprintf(
-					'Could not parse file %1$s: %2$s (line %3$d, column %4$d)',
+					'Could not parse file %1$s.map: %2$s (line %3$d, column %4$d in the concatenated sourcesContent)',
 					$options['file'],
 					$e->getMessage(),
 					$e->getPosition()->getLine(),
--- a/src/MapCodeExtractor.php
+++ b/src/MapCodeExtractor.php
@@ -25,6 +25,11 @@ final class MapCodeExtractor extends JsCode {
 	 * {@inheritdoc}
 	 */
 	public static function fromString( $string, Translations $translations, array $options = [] ) {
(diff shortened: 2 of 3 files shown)


src/MakePotCommand.php
```diff
@@ -554,6 +554,16 @@ protected function extract_strings() {
...
 					]
 				);
+
+				MapCodeExtractor::fromDirectory(
+					$this->source,
+					$translations,
+					[
+						'include'    => $this->include,
+						'exclude'    => $this->exclude,
+						'extensions' => [ 'map' ],
```
swissspidy: `.css.map` files are a thing, so I would make this more explicit by specifying `.js.map`

herregroen: Unfortunately the `getExtension` method used ( http://php.net/manual/en/splfileinfo.getextension.php ) only returns the final extension.

But I can add an additional check in the `MapCodeExtractor` to ensure we're dealing with JS files.

src/MapCodeExtractor.php
```diff
@@ -0,0 +1,60 @@
...
+
+			$functions = new JsFunctionsScanner( $string );
+
+			$functions->enableCommentsExtraction( $options['extractComments'] );
+			$functions->saveGettextFunctions( $translations, $options );
+		} catch ( PeastException $e ) {
+			WP_CLI::debug(
+				sprintf(
+					'Could not parse file %1$s: %2$s (line %3$d, column %4$d)',
+					$options['file'],
```
schlessera: This will point to the `.js` file, not the `.js.map` file, due to the change in line 31. That seems misleading in this case, as the source map failed to parse, not the `.js` file. The `.js` file would probably be a "webpack-ed" JS file, and the line and column references will be completely off.

I think the error message itself might need a clarification, as the line & column probably don't even exist in an actual file in that same location.

schlessera:
What about conflicts in comments between the different files. I guess the `.js` file and the `.js.map` file would either contain the exact same comment (in which case nothing should happen) or the `.js` file has no comment at all, in which the one in the `.js.map` file should be used.

Would there be other scenarios we need to take into account here?

herregroen:
> What about conflicts in comments between the different files. I guess the `.js` file and the `.js.map` file would either contain the exact same comment (in which case nothing should happen) or the `.js` file has no comment at all, in which the one in the `.js.map` file should be used.
> 
> Would there be other scenarios we need to take into account here?

I think at the core this isn't so much a problem in parsing map files but rather a more generic one.

What happens if I have the following PHP file?
```php
// translators: %s is the name of the product.
__( "Add %s", "my-domain" );

// translators: %s is the name of the post type.
__( "Add %s", "my-domain" );
```

Which, in current behaviour would add both comments:
```
public function addExtractedComment($comment)
    {
        if (!in_array($comment, $this->extractedComments, true)) {
            $this->extractedComments[] = $comment;
        }

        return $this;
    }
```

I think that behaviour is fine for map files as well. If the comment is already there then don't add it. If it's not then add it. If there happen to be two different comments there's really no way to automatically determine which one is more correct so the most sensible thing is to add both.

swissspidy:
> What happens if I have the following PHP file?
> Which, in current behaviour would add both comments:

The audit function will print a warning in this case because the string has two different translator comments.

PR #104, opened 2018-10-30
Determine includes and excludes based on scores

herregroen:
Include and exclude matches are based on the strength of the match with:
1. path matches
2. filename matches
3. regex matches

Can probably use improvement but this should cover the base scenarios.

### DIFF
--- a/src/IterableCodeExtractor.php
+++ b/src/IterableCodeExtractor.php
@@ -119,6 +119,7 @@ trait IterableCodeExtractor {
 				)
 			);
 			if ( $base_score === 0 ) {
+				// If the matcher is simply * it gets a score above the implicit score but below 1.
 				$base_score = 0.2;
 			}
 
--- a/src/IterableCodeExtractor.php
+++ b/src/IterableCodeExtractor.php
@@ -95,10 +95,6 @@ trait IterableCodeExtractor {
 	 * @param SplFileInfo $file     File or directory.
 	 * @param array       $matchers List of files and directories to match.
 	 * @return int How strongly the file is matched.
-	 *             0 for not matched.
-	 *             1 for matched by wildcard.
-	 *             2 for matched by direct match on filename.
-	 *             3 for matched by direct match on path.
 	 */
 	protected static function calculateMatchScore( SplFileInfo $file, array $matchers = [] ) {
 		if ( empty( $matchers ) ) {
@@ -106,7 +102,7 @@ trait IterableCodeExtractor {
 		}
 
 		if ( in_array( $file->getBasename(), $matchers, true ) ) {
-			return 2;
+			return 10;
 		}
 
 		// Check for more complex paths, e.g. /some/sub/folder.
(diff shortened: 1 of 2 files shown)


tests/IterableCodeExtractorTest.php
```diff
@@ -70,18 +75,18 @@ public function test_can_exclude_override_wildcard() {
 	}
 
 	public function test_can_exclude_override_matching_directory() {
-		$result     = IterableCodeExtractor::getFilesFromDirectory( self::$base, [ 'foo/bar/*' ], ['excluded'], [ 'php' ] );
+		$result     = IterableCodeExtractor::getFilesFromDirectory( self::$base, [ 'foo/bar/*' ], ['foo/bar/excluded/*'], [ 'php' ] );
```
schlessera: CS: spaces in array notation are not consistent.

src/IterableCodeExtractor.php
```diff
@@ -172,16 +159,16 @@ function ( $file, $key, $iterator ) use ( $include, $exclude, $extensions ) {
 					/** @var RecursiveCallbackFilterIterator $iterator */
 					/** @var SplFileInfo $file */
 
-					if ( static::isExcluded( $file, $exclude ) && ( empty( $include ) || ! static::isIncluded( $file, $include ) ) ) {
-						return false;
+					if ( $iterator->hasChildren() ) {
```
schlessera: We might need to make this smarter, to avoid traversing the entire `node_modules` for example, when there's no rule that "includes" anything related to that.

src/IterableCodeExtractor.php
```diff
@@ -92,64 +92,51 @@ public static function fromDirectory( $dir, Translations $translations, array $o
...
-			$pattern = '/' . str_replace( '__wildcard__', '(.+)', $pattern );
+			$pattern = '(^|/)' . str_replace( '__wildcard__', '(.+)', $pattern );
+
+			$base_score = count(
+				array_filter(
+					explode( '/', $path_or_file ),
+					function ( $component) { return $component !== '*'; }
+				)
+			);
+			if ( $base_score === 0 ) {
```
schlessera: CS: Yoda here, as we use it everywhere else as well.

src/IterableCodeExtractor.php
```diff
@@ -90,61 +90,83 @@ public static function fromDirectory( $dir, Translations $translations, array $o
...
+				0 === strpos( $path_or_file . '/', $root_relative_path )
+			) {
+				return true;
+			}
+
+			$base = current( explode( '*', $path_or_file ) );
 
-			$root_relative_path = str_replace( static::$dir, '', $file->getPathname() );
-			if ( false !== mb_ereg( $pattern, $root_relative_path ) ) {
+			if (
```
schlessera: This needs a comment as well. I'm not 100% sure what the reason is for checking both ways, I assume this is done for being overly conservative. This certainly needs an explanation.

herregroen: Needed for partial wildcards like this:

Given an include wildcard of: `/dir/wp-*` any directory starting with `/dir/wp-` should always be checked.

Whereas for normal wildcards the check is the other way around: `one/two/three/*.js` means that a directory called `one` should be checked to see if it starts with the matcher.

Issue #125, opened 2018-12-19
PHP Warning for make-pot command: Empty needle for strpos()


westonruter:
I've noticed a PHP warning being raised when running the `make-pot` command:

```
Warning: strpos(): Empty needle in phar:///.../wp/vendor/wp-cli/i18n-command/src/IterableCodeExtractor.php on line 175
PHP Warning:  strpos(): Empty needle in phar:///.../wp/vendor/wp-cli/i18n-command/src/IterableCodeExtractor.php on line 175
```

You can see this from a Travis build of the AMP plugin as well: https://travis-ci.org/ampproject/amp-wp/jobs/467674548#L355-L380

swissspidy:
Never ran into this, but it should be an easy fix.

Code in question:

https://github.com/wp-cli/i18n-command/blob/0723170f3201b50bc276a80411b3f84a627d6038/src/IterableCodeExtractor.php#L171-L180

Seems like the following command was run on Travis:

```
wp i18n make-pot . languages/amp-js.pot --include="*.js" --file-comment="*/null/*"
```

@westonruter Out of curiosity, why does the plugin still use make-pot and pot-to-php? Just for WP 4.9 + Gutenberg compatibility? Because with WordPress 5.0 it shouldn't be needed anymore.

westonruter:
@swissspidy Good question.  The plugin is using `make-pot` because it hasn't been updated since the last time you changed it in https://github.com/ampproject/amp-wp/pull/1427 😄 You can see that is where you added the above `--file-comment`.

PR #128, opened 2018-12-26
Extend PotGenerator to improve plural forms output

### LINKED ISSUE #121: POT file generated with this command produces an error with Transifex service related to plural forms
This is a fairly specific issue, but hopefully someone knows something (if the problem is on our end, feel free to close).

We've been using POEdit to (re)generate the POT file for [Pressbooks](https://github.com/pressbooks/pressbooks) and we use [Transifex](https://transifex.com) to manage translations. I have been trying out the wp-cli `make-pot` command and I get this error from the Transifex CLI when trying to upload the resulting POT file:

```
Error received from server: Your source file is not a POT file and the translation file you're using has more than two plurals which is not supported.
```

I ran the command with these options from the plugin root:

```
wp i18n make-pot . languages/_pressbooks.pot --exclude="vendor,symbionts" --headers='{"Report-Msgid-Bugs-To":"https://github.com/pressbooks/pressbooks/"}'
```

I note that in the wp-cli generated file, the second `msgstr` is missing from plural strings:

```diff
#: inc/class-cloner.php:2045 inc/modules/import/wordpress/class-wxr.php:337
#, php-format
msgid "%s part"
msgid_plural "%s parts"
msgstr[0] ""
- msgstr[1] ""
```

Is this expected?

Thanks in advance for any information that might lead to resolution!

swissspidy:
Fixes #121.

Note: Might need some more tests because this adds quite a lot of code from the upstream library.

### DIFF
--- a/src/PotGenerator.php
+++ b/src/PotGenerator.php
@@ -63,7 +63,7 @@ class PotGenerator extends PoGenerator {
 			}
 
 			foreach ( $translation->getReferences() as $reference ) {
-				$lines[] = '#: ' . $reference[0] . ( $reference[1] !== null ? ':' . $reference[1] : null );
+				$lines[] = '#: ' . $reference[0] . ( $reference[1] !== null ? ':' . $reference[1] : '' );
 			}
 
 			if ( $translation->hasFlags() ) {
@@ -91,7 +91,7 @@ class PotGenerator extends PoGenerator {
 			$lines[] = '';
 		}
 
-		return implode("\n", $lines);
+		return implode( "\n", $lines );
 	}
 
 	/**
--- a/src/PotGenerator.php
+++ b/src/PotGenerator.php
@@ -35,7 +35,9 @@ class PotGenerator extends PoGenerator {
 	 * {@parentDoc}.
 	 */
 	public static function toString( Translations $translations, array $options = [] ) {
-		$lines = static::$comments_before_headers + [ 'msgid ""', 'msgstr ""' ];
+		$lines   = static::$comments_before_headers;
+		$lines[] = 'msgid ""';
+		$lines[] = 'msgstr ""';
 
 		$plural_form = $translations->getPluralForms();
 		$plural_size = is_array( $plural_form ) ? ( $plural_form[0] - 1 ) : 1;
--- a/src/PotGenerator.php
+++ b/src/PotGenerator.php
@@ -2,7 +2,7 @@
 
 namespace WP_CLI\I18n;
 
-use Gettext\Generators\Po;
+use Gettext\Generators\Po as PoGenerator;
 use Gettext\Translations;
 
 /**
@@ -11,7 +11,7 @@ use Gettext\Translations;
  * The only difference to the existing PO file generator is that this
  * adds some comments at the very beginning of the file.
  */
-class PotGenerator extends Po {
+class PotGenerator extends PoGenerator {
 	protected static $comments_before_headers = [];
 
 	/**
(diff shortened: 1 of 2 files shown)


greatislander:
@swissspidy I didn't see this— I'll test it with our project and see if it resolves the issue. Thanks!

src/PotGenerator.php
```diff
@@ -35,12 +35,105 @@ public static function setCommentBeforeHeaders( $comment ) {
...
+			}
+
+			if ( $translation->hasExtractedComments() ) {
+				foreach ( $translation->getExtractedComments() as $comment ) {
+					$lines[] = '#. ' . $comment;
+				}
+			}
+
+			foreach ( $translation->getReferences() as $reference ) {
+				$lines[] = '#: ' . $reference[0] . ( $reference[1] !== null ? ':' . $reference[1] : null );
```
schlessera: I'd say `''` is a better no-op here when concatenating:
```suggestion
				$lines[] = '#: ' . $reference[0] . ( $reference[1] !== null ? ':' . $reference[1] : '' );
```

src/PotGenerator.php
```diff
@@ -35,12 +35,105 @@ public static function setCommentBeforeHeaders( $comment ) {
...
+
+		foreach ( $translations as $translation ) {
+			/** @var \Gettext\Translation $translation */
+			if ( $translation->hasComments() ) {
+				foreach ( $translation->getComments() as $comment ) {
+					$lines[] = '# ' . $comment;
+				}
+			}
+
+			if ( $translation->hasExtractedComments() ) {
```
schlessera: This is not really needed, as you'll get an empty array back if not. Iterating over an empty array just does nothing anyway.

src/PotGenerator.php
```diff
@@ -35,12 +35,105 @@ public static function setCommentBeforeHeaders( $comment ) {
...
+				}
+			} else {
+				self::addLines( $lines, $prefix . 'msgstr', $translation->getTranslation() );
+			}
+
+			$lines[] = '';
 		}
 
-		return $result . parent::toString( $translations, $options );
+		return implode("\n", $lines);
```
schlessera: CS:
```suggestion
		return implode( "\n", $lines );
```

src/PotGenerator.php
```diff
@@ -35,12 +35,105 @@ public static function setCommentBeforeHeaders( $comment ) {
...
+
+		foreach ( $translations->getHeaders() as $name => $value ) {
+			$lines[] = sprintf( '"%s: %s\\n"', $name, $value );
+		}
+
+		$lines[] = '';
+
+		foreach ( $translations as $translation ) {
+			/** @var \Gettext\Translation $translation */
+			if ( $translation->hasComments() ) {
```
schlessera: This is not really needed, as you'll get an empty array back if not. Iterating over an empty array just does nothing anyway.

src/PotGenerator.php
```diff
@@ -35,12 +35,105 @@ public static function setCommentBeforeHeaders( $comment ) {
...
 	public static function toString( Translations $translations, array $options = [] ) {
-		$result = '';
+		$lines   = static::$comments_before_headers;
+		$lines[] = 'msgid ""';
+		$lines[] = 'msgstr ""';
 
-		if ( ! empty( static::$comments_before_headers ) ) {
-			$result = implode( "\n", static::$comments_before_headers ) . "\n";
+		$plural_form = $translations->getPluralForms();
+		$plural_size = is_array( $plural_form ) ? ( $plural_form[0] - 1 ) : 1;
```
schlessera: This needs a comment to explain what happens and especially what `$plural_form[0] - 1` means.

schlessera:
The code is not really pretty because there's so much static access going on in the original library, but I guess that can't be easily helped.

connerbw:
Can this be merged soon? I use it.

connerbw:
If this isn't going to be merged soon, can someone sync `121-plural-forms` with `master`? It's ~23 commits behind, there are fixes this branch could benefit from. (For example; #125)  Then I can simply keep using `composer require wp-cli/i18n-command:dev-121-plural-forms`

schlessera:
@swissspidy I'm fine with merging this now. Do you feel good about this as well?

Regarding the tests, I would assume we'll notice when things go wrong because of the Behat tests anyway, even without precise unit tests?

swissspidy:
Sure. We can do some housekeeping here at some point in the future I guess.

schlessera:
@connerbw This is merged into `master` now.

PR #132, opened 2019-01-09
Use extracted text domain when generating JSON files

swissspidy:
https://core.trac.wordpress.org/ticket/45441 got resolved for WordPress 5.0.3, which means we could default to the extracted domain again.

### DIFF
--- a/features/makejson.feature
+++ b/features/makejson.feature
@@ -120,7 +120,47 @@ Feature: Split PO files into JSON files.
       "generator":"WP-CLI
       """
 
-  Scenario: Always uses messages as text domain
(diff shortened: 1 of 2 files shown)


schlessera:
Won't that break for any Core version below 5.0.3?

swissspidy:
That ks correct, yes. We could of course wait for a bit with this change if necessary, but I think eventually we should make it.

schlessera:
I couldn't think of a way to ake this change in a backward-compatible fashion.

I'd suggest adding a note to the documentation of the command to let people know they should use the domain `'messages'` if they intend for the translations to be used with WP < 5.3.

Issue #127, opened 2018-12-19
Improve path handling when generating JSON files


swissspidy:
I think we need to make the `make-json` command a bit smarter.

Given the following folder structure:

```
- my-plugin/
  - js/
    blocks.js
    - src/
      blocks.js
  - languages/
    my-plugin.pot
    my-plugin-de_DE.po
    my-plugin-de_DE.mo
```

Let's say there are some translatable strings in `my-plugin/js/src/blocks.js` and now we extract these somehow.

Scenario A: We use `babel-plugin-makepot` for string extraction. The POT and PO files will only have `my-plugin/js/src/blocks.js` as a file reference.

Scenario B: We use WP-CLI. All files are taken into consideration for string extraction. The POT and PO files will have both `my-plugin/js/src/blocks.js` and `my-plugin/js/blocks.js` as file references.

Scenario C: We use WP-CLI for string extraction, but we ignore the `src` folder. Not ideal, because this way you lose the context the original source file, but it works. The POT and PO files will have `my-plugin/js/blocks.js` as a file reference.

In each scenario, we run `wp i18n make-json`.

In WordPress, we enqueue the script `my-plugin/js/blocks.js`. This means WordPress looks for JSON files with the MD5 hash of `js/blocks.js` in the file name. This is where we run into problems.

Scenario A: WP-CLI successfully creates 1 JSON file, but uses the MD5 hash of `js/src/blocks.js` for creating the file name. There's obviously a mismatch because WordPress expects a different file name.

Scenario B: WP-CLI creates 2 JSON files, one for `js/src/blocks.js` and one for `js/blocks.js`. The first one doesn't work, but the second is what WordPress expects.

Scenario C: WP-CLI creates 1 JSON file with the MD5 hash of `js/blocks.js`. The same as WordPress expects.

How could we fix this?

Scenario A: Don't use the MD5 in the file name, but the script handle instead. This would need to be passed as an argument, obviously. However, a plugin can have lots of script handles, so this is gonna be a bit complicated. Alternatively, we'd need a way to tell WP-CLI to use `js/blocks.js` for the MD5 hash instead of `js/src/blocks.js`.

Scenario B: Add a way to tell WP-CLI to ignore certain paths when generating the JSON files.

Scenario C: Nothing to do there.

PR #133, opened 2019-01-09
Improve language handling when creating JSON files

swissspidy:
Turns out `\Gettext\Translations::setLanguage()` throws an exception when trying to pass a language that's not	in the list provided by https://packagist.org/packages/gettext/languages (CLDR).

That means it fails when passing `de_DE`, because it expects `de-DE`.

This PR does two things:

* Replaces underscore with hyphens
* Catches exceptions and falls back to "en" in that case.

### DIFF
--- a/features/makejson.feature
+++ b/features/makejson.feature
@@ -514,7 +514,7 @@ Feature: Split PO files into JSON files.
               "messages": {
                   "": {
                       "domain": "messages",
-                      "lang": "de_DE",
+                      "lang": "de-DE",
                       "plural-forms": "nplurals=2; plural=(n != 1);"
                   },
                   "Foo Plugin": [
@@ -522,4 +522,41 @@ Feature: Split PO files into JSON files.
                   ]
               }
           }
+      """
+
+  Scenario: Should fall back to English for invalid locales.
+    Given an empty foo-plugin directory
+    And a foo-plugin/foo-plugin-invalid.po file:
+      """
+      # Copyright (C) 2018 Foo Plugin
+      # This file is distributed under the same license as the Foo Plugin package.
+      msgid ""
+      msgstr ""
+      "Project-Id-Version: Foo Plugin\n"
+      "Report-Msgid-Bugs-To: https://wordpress.org/support/plugin/foo-plugin\n"
+      "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+      "Language-Team: LANGUAGE <LL@li.org>\n"
+      "Language: invalid\n"
+      "MIME-Version: 1.0\n"
+      "Content-Type: text/plain; charset=UTF-8\n"
+      "Content-Transfer-Encoding: 8bit\n"
+      "POT-Creation-Date: 2018-05-02T22:06:24+00:00\n"
+      "PO-Revision-Date: 2018-05-02T22:06:24+00:00\n"
+      "X-Domain: foo-plugin\n"
+      "Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+      #: foo-plugin.js:15
+      msgid "Foo Plugin"
+      msgstr "Foo Plugin"
+      """
+
+    When I run `wp i18n make-json foo-plugin`
+    Then STDOUT should contain:
+      """
+      Success: Created 1 file.
+      """
+    And the return code should be 0
+    And the foo-plugin/foo-plugin-invalid-56746e49c6485323d16a717754b7447e.json file should contain:
+      """
+      "lang":"en"
       """
--- a/src/MakeJsonCommand.php
+++ b/src/MakeJsonCommand.php
@@ -155,7 +155,14 @@ class MakeJsonCommand extends WP_CLI_Command {
 					$mapping[ $source ] = new Translations();
 					// See https://core.trac.wordpress.org/ticket/45441
 					//$mapping[ $source ]->setDomain( $translations->getDomain() );
-					$mapping[ $source ]->setLanguage( $translations->getLanguage() );
+
+					try {
+						$mapping[ $source ]->setLanguage( str_replace( '_', '-', $translations->getLanguage() ) );
+					} catch ( \InvalidArgumentException $e ) {
+						// The locale had an invalid format.
+						$mapping[ $source ]->setLanguage( 'en' );
+					}
+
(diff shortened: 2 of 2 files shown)


alialaa:
Hello, I still get this error even after replacing underscores with dashes. When I run the make-json command, I get absolutely nothing. No errors, no success messages and no files created. I went to `users/myuser/.wp-cli/packages/vendor/wp-cli/i18n-command` to debug and I var_dumped the error in [this handler](https://github.com/wp-cli/i18n-command/blob/master/src/MakeJsonCommand.php#L161). And it says: `string(33) "The language "de-DE" is not valid"` apparently this error is thrown no matter what string is passed there. So even `$mapping[ $source ]->setLanguage( 'en' );` in the error handler will throw an error. I tried hardcoding different strings and none worked.

swissspidy:
@alialaa Are you using the latest version of the command that includes the fixes from this PR?

alialaa:
I guess so I ran `php -d memory_limit=512M "$(which wp)" package install git@github.com:wp-cli/i18n-command.git` and I am trying the code on [this repository.](https://github.com/swissspidy/gutenberg-i18n-block)

alialaa:
Also in `users/myuser/.wp-cli/packages/vendor/wp-cli/i18n-command/src/MakeJsonCommand` I have this line `$mapping[ $source ]->setLanguage( str_replace( '_', '-', $translations->getLanguage() ) );`. I guess that means I am having the latest version?

swissspidy:
Ah yes, that looks correct.

Looks like the issue is that the language list (https://github.com/mlocati/cldr-to-gettext-plural-rules/blob/master/src/cldr-data/main/en-US/languages.json) includes `de` but not `de-DE`. However, it does include `en-US` _and_ `en`.

The upstream library we use leverages the language information solely for the purpose of adding plural forms: https://github.com/oscarotero/Gettext/blob/545f716f08338030d2d0d35a1835d04a9d4f18ed/src/Translations.php#L352-L361

However, we already do that ourselves anyway: https://github.com/wp-cli/i18n-command/blob/af14c1650751c0ffee2b06e7a5e58ae0abfca5fe/src/MakeJsonCommand.php#L168-L173

So I guess we can just silently ignore that exception...

alialaa:
The thing is if I hard code 'en' like so in the try catch block, I also get `The languane 'en' is not valid`:
```
try {
   var_dump(Language::getAll());
   $mapping[ $source ]->setLanguage( 'en' );
} catch ( \InvalidArgumentException $e ) {
   // The locale had an invalid format.
   var_dump($e->getMessage());
   $mapping[ $source ]->setLanguage( 'en' );
}
```
Also not sure if this is related but Language::getAll() that I am dumping returns an empty array. Languages is from `use Gettext\Languages\Language;`

swissspidy:
Weird, I'm pretty sure that worked in tests. But not really a problem. With my suggestion (created #142 for it) we don't need that list anyway.

alialaa:
So I can just remove that whole try, catch block if I need to test it now, correct?

swissspidy:
You could certainly do that for testing, yes, but that's not really the solution. We definitely should set the language for the translations.

The correct solution would probably be more like this:

```php
try {
	$mapping[ $source ]->setLanguage( $translations->getLanguage() );
} catch ( \InvalidArgumentException $e ) {
	// Do nothing.
}
```

Or perhaps

```php
$mapping[ $source ]->setHeader( \Gettext\Translations::HEADER_LANGUAGE, $translations->getLanguage() );
```

alialaa:
Thanks, I will wait for your fix. I just tried this and it worked but I have some notes that maybe could be helpful to you:

1. The js strings are removed from the PO files after generating the JSON. I know this is intended but what if I wanted to update the PO file later, how will I get the old values that are now gone?
2. Also a weird thing happened with me. Not sure if it's intentional but I had multiple json files generated. 1 for each JS source file in the PO file.
3. Also the output file should be called {pluginName}-{locale}-{handler}.json (where handler is the first arg in wp_set_script_translations) in order for the translations to work. So maybe you can use that instead of the hash?

Thanks for your help!

swissspidy:
@alialaa 

1. Use the `--no-purge` flag to skip this, as [per the documentation](https://github.com/wp-cli/i18n-command#wp-i18n-make-json).
2. That is intentional. Please read the [documentation](https://github.com/wp-cli/i18n-command#wp-i18n-make-json) and [this longer explanation](https://pascalbirchler.com/internationalization-in-wordpress-5-0/).
3. No. Or to be more precise: _It depends_. `{pluginName}-{locale}-{handle}.json` and `{pluginName}-{locale}-{hash}.json` **both work**. The problem is, WP-CLI does not know your script handle, that's why. Neither does WordPress.org. That's why  we need to use the hash variant. However, for WP-CLI we might need to figure out how you could inform WP-CLI of the script handles in order to use the other variant. See #127 for this.

alialaa:
Thanks a lot :)

michaelzangl:
I enocuntered the same issue. This is what I do to fix the wp-cli.phar

* Remove `Symfony\Component\Finder\`
* Replace `$mapping[ $source ]->setLanguage( $translations->getLanguage() );` by `$mapping[ $source ]->setHeader( \\Gettext\\Translations::HEADER_LANGUAGE, $translations->getLanguage() );`


```
php -c -r "$phar = new Phar('.....wp-cli.phar'); $phar['vendor/wp-cli/i18n-command/src/MakeJsonCommand.php'] = str_replace( ['Symfony\\Component\\Finder\\\\', '$mapping[ $source ]->setLanguage( $translations->getLanguage() );'], ['', '$mapping[ $source ]->setHeader( \\Gettext\\Translations::HEADER_LANGUAGE, $translations->getLanguage() );'], $phar['vendor/wp-cli/i18n-command/src/MakeJsonCommand.php']->getContent());
```

cbratschi:
Which WP-CLI version will include this fix?

swissspidy:
@cbratschi The next one? :-)

Once #150 is merged, you can install the latest version of this package over what's included in WP-CLI by running this:

```
wp package install git@github.com:wp-cli/i18n-command.git
```

This way you don't have to wait for the next WP-CLI release.

cbratschi:
@swissspidy Thanks a lot, it works now fine.

Issue #117, opened 2018-12-06
Probable bug: IterableCodeExtractor::calculateMatchScore() - mbstring vs PCRE


jrfnl:
Note: This is purely based on a code-review, so I have no test case currently (though I can make one up if needs be).

In the `IterableCodeExtractor::calculateMatchScore()` method there are two things which caught my attention:
1. The `preg_quote()` function is used to escape arbitrary strings, while `mb_ereg()` is used for the regex matching.
    This is problematic as `preg_quote()` is part of the PCRE extension, which uses the PCRE regex engine, while `mb_ereg()` is part of the MBString extension, which uses the Oniguruma regex engine.
   These engines are not 100% compatible, so in effect you could be escaping too much/too little by using `preg_quote()`, such as the delimiter (used in PCRE, not used in MBString).
2. The `mb_ereg()` function is used without the [`mb_internal_encoding`](http://php.net/manual/en/function.mb-internal-encoding.php) or the [`mb_regex_encoding`](http://php.net/manual/en/function.mb-regex-encoding.php) being set.
    Whether `mb_internal_encoding` is needed may depend on where the input comes from, however `mb_regex_encoding` should most definitely be set.
    The default `mb_regex_encoding` is `EUC-JP` in PHP 5.4 and 5.5 and only became `UTF-8` in PHP 5.6, though as other code may have also called this function, the encoding being the default can not be relied upon and it should be set before using `mb_ereg()`.
    Note: `mb_internal_encoding` was deprecated in PHP 5.6 in favour of `default_encoding`, so you need a compatibility layer here.

PR #156, opened 2019-04-17
Implement CS checking based on the `WP_CLI_CS` ruleset

### LINKED ISSUE #158: Adopt and enforce new `WP_CLI_CS` standard
We have a new PHPCS standard for WP-CLI called `WPCliCS` (props @jrfnl). It is part of the `wp-cli/wp-cli-tests` package starting with version v2.1.0.

To adopt & enforce this new standard, the following actions need to be taken for this repository:

- [x] Create a PR that adds a custom ruleset `phpcs.xml.dist` to the repository

    - Add `phpcs.xml.dist` file
    - Adapt `.distignore` to ignore `phpcs.xml.dist` & `phpunit.xml.dist`
    - Adapt `.gitignore` to ignore `phpunit.xml`, `phpcs.xml` & `.phpcs.xml`
    - Require version `^2.1` of the `wp-cli/wp-cli-tests` as a dev dependency
- [x] Make any required changes to the code that fail the checks from the above ruleset in separate PRs
- [ ] Merge thre ruleset once all required changes have been processed and merged

A sample PR for a simple repository can be seen here: https://github.com/wp-cli/maintenance-mode-command/pull/3/files

Related https://github.com/wp-cli/wp-cli/issues/5179

jrfnl:
### Composer: use wp-cli-tests ^2.1

... as that version contains the new WPCliCS PHPCS standard.

### .gitignore/.distignore: ignore phpcs/phpunit config files

* `.distignore`: no need to distribute the phpcs/phpunit config files.
* `.gitignore`: Allow devs to overwrite config files for PHPUnit and PHPCS
    - The `.dist` files should be committed. However, for their personal use, devs can overrule those files with versions without `.dist`.
        Those personal versions should never be committed.

As a side-note: for PHPCS, having a personal version while still using the original `.dist` file is made very easy, as you can just import the `.dist` file as a starting point, like so:
```xml
<?xml version="1.0"?>
<ruleset name="WP-CLI">
    <rule ref="./phpcs.xml.dist"/>
</ruleset>
```

### PHPCS: add a ruleset for this project

The file is completely documented in-line.

Fixes #158 

Related https://github.com/wp-cli/wp-cli/issues/5179

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     },
     "require-dev": {
         "wp-cli/scaffold-command": "^1.2 || ^2",
-        "wp-cli/wp-cli-tests": "^2.0.7"
+        "wp-cli/wp-cli-tests": "^2.1"
     },
     "config": {
         "process-timeout": 7200,
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -1,0 +1,58 @@
+<?xml version="1.0"?>
+<ruleset name="WP-CLI-i18n">
+	<description>Custom ruleset for WP-CLI I18n-command</description>
+
+	<!--
+	#############################################################################
+	COMMAND LINE ARGUMENTS
+	For help understanding this file: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml
+	For help using PHPCS: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Usage
+	#############################################################################
+	-->
+
+	<!-- What to scan. -->
+	<file>.</file>
+
+	<!-- Ignoring select files/folders.
+		 https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-files-and-folders -->
+	<exclude-pattern>*/tests/data/*</exclude-pattern>
+
+	<!-- Show progress. -->
+	<arg value="p"/>
+
+	<!-- Strip the filepaths down to the relevant bit. -->
+	<arg name="basepath" value="./"/>
+
+	<!-- Check up to 8 files simultaneously. -->
+	<arg name="parallel" value="8"/>
+
+	<!--
+	#############################################################################
+	USE THE WP_CLI_CS RULESET
+	#############################################################################
+	-->
+
+	<rule ref="WP_CLI_CS"/>
+
+	<!--
+	#############################################################################
+	PROJECT SPECIFIC CONFIGURATION FOR SNIFFS
+	#############################################################################
+	-->
+
+	<!-- For help understanding the `testVersion` configuration setting:
+		 https://github.com/PHPCompatibility/PHPCompatibility#sniffing-your-code-for-compatibility-with-specific-php-versions -->
+	<config name="testVersion" value="5.4-"/>
+
+	<!-- Verify that everything in the global namespace is either namespaced or prefixed.
+		 See: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#naming-conventions-prefix-everything-in-the-global-namespace -->
+	<rule ref="WordPress.NamingConventions.PrefixAllGlobals">
+		<properties>
+			<property name="prefixes" type="array">
+				<element value="WP_CLI\I18n"/><!-- Namespaces. -->
+				<element value="wpcli_i18n"/><!-- Global variables and such. -->
+			</property>
+		</properties>
+	</rule>
+
+</ruleset>
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -55,4 +55,24 @@
 		</properties>
 	</rule>
 
+	<!-- Whitelist method names for classes that extend a PSR-2 package. -->
+	<rule ref="WordPress.NamingConventions.ValidFunctionName.MethodNameInvalid">
+		<exclude-pattern>*/src/IterableCodeExtractor\.php$</exclude-pattern>
+	</rule>
+
+	<!-- Whitelist variable names for classes that extend a PSR-2 package. -->
+	<rule ref="WordPress.NamingConventions.ValidVariableName">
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
     },
     "require-dev": {
         "wp-cli/scaffold-command": "^1.2 || ^2",
-        "wp-cli/wp-cli-tests": "^2.0.7"
+        "wp-cli/wp-cli-tests": "^2.0.14"
```
schlessera:
```suggestion
        "wp-cli/wp-cli-tests": "^2.1"
```

jrfnl: That was a large part of the reason that this is a draft PR :stuck_out_tongue_winking_eye:

jrfnl:
Remaining issues list: 
[20190417-phpcs-result-i18n.txt](https://github.com/wp-cli/i18n-command/files/3090358/20190417-phpcs-result-i18n.txt)

Once these issues have been resolved, either by fixing them, whitelisting individual issues or pinging me to adjust the ruleset, this PR will need to be rebased.

After that it can be marked `Ready for review` and the build should go green.

schlessera:
@jrfnl I pushed changes into this PR, it should now pass PHPCS.

phpcs.xml.dist
```diff
@@ -0,0 +1,78 @@
...
+			</property>
+		</properties>
+	</rule>
+
+	<!-- Whitelist method names for classes that extend a PSR-2 package. -->
+	<rule ref="WordPress.NamingConventions.ValidFunctionName.MethodNameInvalid">
+		<exclude-pattern>*/src/IterableCodeExtractor\.php$</exclude-pattern>
+	</rule>
+
+	<!-- Whitelist variable names for classes that extend a PSR-2 package. -->
```
jrfnl: It would be better to use the specific configuration for this sniff to allow for these.
See: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#mixed-case-property-name-exceptions

swissspidy:
Whoops 🙃 Sorry about that.

phpcs.xml.dist
schlessera: @jrfnl There was actually three different ones that were triggered, so I didn't bother listing them individually anymore.

src/IterableCodeExtractor.php
```diff
@@ -111,7 +111,7 @@ protected static function calculateMatchScore( SplFileInfo $file, array $matcher
 		$root_relative_path = str_replace( static::$dir, '', $file->getPathname() );
 
 		foreach ( $matchers as $path_or_file ) {
-			$pattern = preg_quote( str_replace( '*', '__wildcard__', $path_or_file ) );
+			$pattern = preg_quote( str_replace( '*', '__wildcard__', $path_or_file ), '/' );
```
jrfnl: This disregards the bug reported in #117 - MBString regexes don't use delimiters and using `preg_quote()` on a regex to be passed to one of the MBString regex functions is risky and potentially buggy.

schlessera:
> @schlessera Thanks. I have a couple of remarks, but it seems that @swissspidy already merged this....

Just post the remarks here, and I'll follow up in a separate PR as needed.

jrfnl:
I just have ;-)

src/IterableCodeExtractor.php
schlessera: @jrfnl Noted. We have the bug report open, so the issue won't be forgotten. I assume it is a bigger refactor to fix this, though. We're not worse off for now with the above code, as the bug was detected and recorded for a later fix.

schlessera:
I replied to them. Do you agree with my assessments, or is there something I did not consider?

phpcs.xml.dist
jrfnl: The problem with that is that now newly introduced variable name violations in these files will no longer be reported, which is why I suggested using the property.

jrfnl:
@schlessera I agree with the one, not so much the other ;-)

schlessera:
The `mb_ereg/preg_quote` mismatch, I assume.

I'm happy to look into actually fixing the bug for this release if I end up having time once the blocking bugs are fixed. This one hasn't bitten anyone yet and doesn't block the release, so it is low priority for me right now.

phpcs.xml.dist
schlessera: Oh, I'm sorry, I should have actually read the link you posted. So instead of whitelisting against the rule, I should whitelist against the individual variable names.

Yes, I agree that that makes more sense, I'll push a subsequent PR.

schlessera: Hmm, that only seems to work for properties, not variables. Am I missing a specific notation?
Here's what seems to work for the properties:
Image 2019-04-24 at 5 57 28 PM
This leaves me with the following errors:
Image 2019-04-24 at 5 58 10 PM

jrfnl: Having a look now at the files in question. This is what I see:

1. `src/JSFunctionsScanner.php`: I see no reason why the `$extractComments` property hasn't been renamed. It is a `private` property, so independent of the parent class. Why should this not comply ?
2. `src/MapCodeExtractor.php`: `$mapObject` is a function local variable, so it is perfectly safe to rename it. Why should this not comply ?
3. `src/MapCodeExtractor.php`: `sourcesContent` is a property and can be whitelisted via the above.
4. `src/PhpFunctionsScanner.php`: `$extractedComment` is a function local variable, so it is perfectly safe to rename it. Why should this not comply ?

So basically this would be the ruleset config. Everything else should just be renamed:
```xml
	<rule ref="WordPress.NamingConventions.ValidVariableName">
		<properties>
			<property name="customPropertiesWhitelist" type="array">
				<element value="sourcesContent"/>
			</property>
		</properties>
	</rule>

```

schlessera: You're right, I did these changes in https://github.com/wp-cli/i18n-command/pull/161

PR #160, opened 2019-04-22
make pot command: Allow .jsx files.

michaelzangl:
Currently, the js file scanner only scans for .js files.
In react projects, some developers name their react-files ending in .jsx
Since Gutenberg is using react, it won't hurt allowing those files to be parsed.

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -1638,6 +1638,48 @@ Feature: Generate a POT file of a WordPress project
       msgid "wrong-domain"
       """
 
+  Scenario: Parse .js and .jsx files for javascript translations
+    Given an empty foo-plugin directory
+    And a foo-plugin/foo-plugin.php file:
+      """
+      <?php
+      /**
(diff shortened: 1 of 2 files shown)


swissspidy:
Thanks for your PR. Would you be able to extend the current set of tests to make sure strings from .JSX files are properly extracted?

michaelzangl:
I can do that, but it will take a few days.

swissspidy:
For reference, the current test for `.js` string extraction is here:

https://github.com/wp-cli/i18n-command/blob/4d3e4857765c1d6bfc4d95d47797f7a97203e517/features/makepot.feature#L1501-L1639

This can be duplicated and adjusted with some JSX-only code.

michaelzangl:
I added a different test. I don't think we need to test for all of the syntax in the .jsx. The feature here is the detection of the correct file extension. So the test is, that .js and .jsx files are both detected as JavaScript files and the JavaScript parser is used for both of them.

PR #161, opened 2019-04-24
Fine-tune property and variable name whitelisting

schlessera:
Late addition to #156

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -60,11 +60,13 @@
 		<exclude-pattern>*/src/IterableCodeExtractor\.php$</exclude-pattern>
 	</rule>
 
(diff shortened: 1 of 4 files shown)


phpcs.xml.dist
```diff
@@ -60,11 +60,13 @@
 		<exclude-pattern>*/src/IterableCodeExtractor\.php$</exclude-pattern>
 	</rule>
 
-	<!-- Whitelist variable names for classes that extend a PSR-2 package. -->
+	<!-- Whitelist property name for a class that extends a PSR-2 package. -->
```
jrfnl: This documentation may need some tuning as well.

AFAICS the property comes from an object which is `json_decode()`d and not necessarily related to the parent class this class extends.

PR #164, opened 2019-04-28
Workaround for parsing dynamic import.

michaelzangl:
Peast does not support dynamic import syntax
https://github.com/mck89/peast/issues/7

This is a workaround for dynamic imports (to silently ignore them, they don't affect i18n generation).

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -1638,6 +1638,34 @@ Feature: Generate a POT file of a WordPress project
       msgid "wrong-domain"
       """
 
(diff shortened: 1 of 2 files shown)


swissspidy:
For the record, the [upstream issue](https://github.com/mck89/peast/issues/7) was closed with the following explanation:

> As you can see [here](https://github.com/tc39/proposals) dynamic import is still a proposal, it's not in the official ECMAScript specification yet and that's why it's not supported at the moment. I know some browsers already implement it, but i want this project to be conformant to the specs since proposals can change at any time until they get included.

Issue #171, opened 2019-07-15
Avoid large language file diffs on different filesystems by sorting translatable files


l3ku:
## Feature Request

 

- [X] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

When working on translations with developers that have different operating systems, 
the `wp i18n makepot` command causes large diffs that make reviewing language changes more challenging. The reason for this is that the command relies on the order the files are returned by the filesystem, and different filesystem implementations on the different operating systems of developers return the files in an arbitrary order.

A verbose example of the situation:

1. The first developer on e.g. Mac OS X runs `wp i18n makepot` on his/her machine, creating the source code files are traversed in the order returned by the OS
2. The first developer commits to git
3. The second developer on Ubuntu clones the git repo, creates a new branch and adds a single translatable line
4. The second developer regenerates the `.pot` file with `wp i18n makepot`, adds a translation with Poedit and pushes his/her branch

The strings in the`.pot` file will be reordered totally, and as a result also the actual translations in the `.po` files will follow this same new order. Reviewing the translation will be difficult.

**Describe the solution you'd like**

Always sort the translatable file paths in alphabetic order so the file order will be identical on all machines.

swissspidy:
Thanks for raising this!

Sounds like a reasonable thing to do.

PR #173, opened 2019-07-19
Sort translation files by name to ensure same order on all systems

l3ku:
See wp-cli/i18n-command#171 for a better explanation on why to sort the translation files.

### DIFF
--- a/src/IterableCodeExtractor.php
+++ b/src/IterableCodeExtractor.php
@@ -234,6 +234,8 @@ trait IterableCodeExtractor {
 			$filtered_files[] = Utils\normalize_path( $file->getPathname() );
 		}
 
+		sort( $filtered_files, SORT_NATURAL | SORT_FLAG_CASE );
+
 		return $filtered_files;
 	}
 }
--- a/tests/IterableCodeExtractorTest.php
+++ b/tests/IterableCodeExtractorTest.php
@@ -110,4 +110,17 @@ class IterableCodeExtractorTest extends PHPUnit_Framework_TestCase {
 		$expected = static::$base . 'foo/bar/excluded/ignored.js';
 		$this->assertContains( $expected, $result );
 	}
+
+	public function test_can_return_all_directory_files_sorted() {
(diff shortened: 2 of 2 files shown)


swissspidy:
Would you mind adding some tests demonstrating that this works? :-)

l3ku:
I have now included a test case into `/tests/IterableCodeExtractorTest.php`. I did not add any new test data files into `/tests/data`, do you think that validating that all of the already existing files are returned in the intended order is sufficient or should I make the test more extensive?

src/IterableCodeExtractor.php
```diff
@@ -233,7 +233,7 @@ function ( $file, $key, $iterator ) use ( $include, $exclude, $extensions ) {
 
 			$filtered_files[] = Utils\normalize_path( $file->getPathname() );
 		}
-
+		sort( $filtered_files );
```
swissspidy: Should this use the `SORT_NATURAL` flag?

l3ku: Or then perhaps [natsort](https://www.php.net/manual/en/function.natsort.php) could be used? As long as the sorting behaves the same way on all systems without taking e.g. different system locales into consideration, then https://github.com/wp-cli/i18n-command/issues/171 will be resolved regardless of whether the files are sorted in natural order or not. Would sorting in natural order provide some additional benefit?

swissspidy: `SORT_NATURAL` or `natsort` don't take locales into account, so that should not be a problem.

And yes, this PR seems to already solve #171. I was just wondering whether it would make sense to have more natural sorting. Examples from the docs page:

```
Standard sorting
Array
(
    [3] => img1.png
    [1] => img10.png
    [0] => img12.png
    [2] => img2.png
)

Natural order sorting
Array
(
    [3] => img1.png
    [2] => img2.png
    [1] => img10.png
    [0] => img12.png
)
```

The latter makes more sense to me there :-)

l3ku: Added `SORT_NATURAL` and `SORT_FLAG_CASE` flags (PHP>=5.4.0) to `sort` so that the sorting order is more natural :). Did not use `natsort` because it maintains key-value associations which is not necessary when having numerical indices.

Issue #172, opened 2019-07-17
Vendor folder ignored


BartvanS:
Referece: #84

- [X] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [X] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**
When i used the command 'make-pot' I found out that folder such as vendor and git are ignored. To solve the issue i found this thread https://github.com/wp-cli/i18n-command/issues/84.

scenario:
I have a test theme in my wordpress. In this theme there is the stye.css, a test.php with the content `<?php __('test', 'test');`. There is also a copy of this file in a folder named "vendor"

if i use the `--include` parameter with the test.php file, everything works fine. When i add `--include=vendor`, vendor is stile excluded and i create a not complete pot file.

my run script: `wp i18n make-pot ./ ./test.pot --ignore-domain --include=vendor`

**Describe what you would expect as the correct outcome**

- when running command without `--include=vendor`:

```
#. Theme Name of the theme
#. Description of the theme
msgid "Het Online Recept"
msgstr ""

#. Theme URI of the theme
#. Author URI of the theme
msgid "https://bluemammoth.nl/"
msgstr ""

#. Author of the theme
msgid "Blue Mammoth"
msgstr ""

#: test.php:3
msgid "test"
msgstr ""


- When i run with `--include=vendor`:
#. Theme Name of the theme
#. Description of the theme
msgid "Het Online Recept"
msgstr ""

#. Theme URI of the theme
#. Author URI of the theme
msgid "https://bluemammoth.nl/"
msgstr ""

#. Author of the theme
msgid "Blue Mammoth"
msgstr ""
```
(Paste the output of "wp cli info" into this box)

```OS:     Windows NT 10.0 build 17763 (Windows 10) i586
Shell:  C:\Program Files\Git\usr\bin\bash.exe
PHP binary:     C:/xampp/php/php
PHP version:    7.2.12
php.ini used:   C:\xampp\php\php.ini
WP-CLI root dir:        C:\Users\Stage\AppData\Roaming\Composer\vendor\wp-cli\wp-cli
WP-CLI vendor dir:      C:\Users\Stage\AppData\Roaming\Composer\vendor
WP_CLI phar path:
WP-CLI packages dir:
WP-CLI global config:
WP-CLI project config:
[...]

PR #175, opened 2019-07-23
Investigate include/exclude oddities

swissspidy:
I've tried replicating #172 with the similar setup:

```
- test.php
- vendor/
- vendor/test.php
```

I ran:

```
vendor/bin/wp i18n make-pot /tmp/wp-i18n-test --debug --include=/vendor/test.php
```

And the `vendor/test.php` file was **not** included as expected.

Then I added 763941ad420d9448c51de738a9a40863d87cf362 and the file was now **included** as expected.

However, I failed at verifying this with a PHPUnit test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 .DS_Store
 wp-cli.local.yml
-node_modules/
-vendor/
+/node_modules
+/vendor
 *.zip
 *.tar.gz
 *.swp
--- a/src/IterableCodeExtractor.php
+++ b/src/IterableCodeExtractor.php
@@ -159,6 +159,7 @@ trait IterableCodeExtractor {
 
 		/** @var string $root_relative_path */
 		$root_relative_path = str_replace( static::$dir, '', $dir->getPathname() );
+		$root_relative_path = ltrim( $root_relative_path, '/' );
 
 		foreach ( $matchers as $path_or_file ) {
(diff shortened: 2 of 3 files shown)


schlessera:
I looked into the tests, and for the failing test, the function to extract the files returns an empty array. I'll take a closer look at the code now.

schlessera:
From what I can see on my system, this change does nothing, the strings that get passed into that method never contain a leading slash for the tests you are doing...

schlessera:
@swissspidy Can you take a look at the code change I made to validate?

What I found is that the logic that was in place include directories that had files that might match ... but excluded the actual files that would have matched.

schlessera:
Okay, it's obviously not enough to have the unit tests pass...

BartvanS:
I updated my wp-cli and wp cli i18n. Sadly if i recreate this scenario i still only get the translations from the test.php and not the vendor/test.php.
$ wp cli  version -> 2.3.0
I installed the latest version of wp i18n with: 
$ wp package install git@github.com:wp-cli/i18n-command.git

swissspidy:
@BartvanS What command are you running?

BartvanS:
wp i18n make-pot ./ ./test.pot --ignore-domain --include=/vendor/test.php

schlessera:
@BartvanS Do you maybe have the `wp-cli/i18n-command` package installed as an external package as well? You can check this via `wp package list`.

PR #183, opened 2019-10-18
Add `--debug` output for the file being processed.

ocean90:
For debugging purposes it would be helpful to know what file is currently processed.

[Initially suggested](https://wordpress.slack.com/archives/C02RP4T41/p1571379966097800?thread_ts=1571300642.061700&cid=C02RP4T41) by @schlessera.

### DIFF
--- a/src/JsCodeExtractor.php
+++ b/src/JsCodeExtractor.php
@@ -28,6 +28,13 @@ final class JsCodeExtractor extends JsCode {
 		try {
 			$options += static::$options;
 
+			WP_CLI::debug(
+				sprintf(
+					'Parsing file %s',
+					$options['file']
+				)
+			);
+
 			$functions = new JsFunctionsScanner( $string );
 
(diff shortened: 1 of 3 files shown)


src/JsCodeExtractor.php
```diff
@@ -28,6 +28,13 @@ public static function fromString( $string, Translations $translations, array $o
 		try {
 			$options += static::$options;
 
+			WP_CLI::debug(
```
schlessera: This takes up a lot of visual space for a debugging statement.

I'd prefer the following notation instead:
```php
WP_CLI::debug( "Parsing file {$options['file']}" );
```

schlessera:
@ocean90 Using `sprintf()` and multiple lines here is overkill and detracts from the actual logic. I'd prefer each of these to be single lines instead.

schlessera:
All failures unrelated (and waiting for https://github.com/oscarotero/Gettext/issues/231 to get solved), so merging as-is.

PR #184, opened 2019-10-18
Bail early for files which don't match the requested extensions

ocean90:
We shouldn't calculate match scores for files which will be ignored.

### DIFF
--- a/src/IterableCodeExtractor.php
+++ b/src/IterableCodeExtractor.php
@@ -205,7 +205,7 @@ trait IterableCodeExtractor {
 					/** @var RecursiveCallbackFilterIterator $iterator */
 					/** @var SplFileInfo $file */
 
-					if ( $file->isFile() && ! in_array( $file->getExtension(), $extensions, true ) ) {
+					if ( ! $file->isFile() || ! in_array( $file->getExtension(), $extensions, true ) ) {
 						return false;
 					}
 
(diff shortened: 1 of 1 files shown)


src/IterableCodeExtractor.php
```diff
@@ -223,7 +227,7 @@ function ( $file, $key, $iterator ) use ( $include, $exclude, $extensions ) {
 						return static::containsMatchingChildren( $file, $include );
 					}
 
-					return ( ( $inclusion_score >= $exclusion_score ) && $file->isFile() && in_array( $file->getExtension(), $extensions, true ) );
+					return $inclusion_score >= $exclusion_score;
```
schlessera: This has now changed the logic, as we would now return a score for non-Files as well.
```suggestion
					return $inclusion_score >= $exclusion_score && $file->isFile();
```
@swissspidy Am I missing something here?

swissspidy: Nope, you're right.

Could we perhaps change the condition above to:

```php
if ( ! $file->isFile() || ( $file->isFile() && ! in_array( $file->getExtension(), $extensions, true ) ) ) {
	return false;
}
```

instead?

schlessera: The second `$file->isFile()` is redundant. If the first one succeeds, it stops there, if it fails, then `isFile()` is guaranteed to be true.

schlessera: So:
```php
if ( ! $file->isFile() || ! in_array( $file->getExtension(), $extensions, true ) ) {
	return false;
}
```

schlessera:
This change seems to cause a lot of breakage.

PR #198, opened 2019-12-11
Fix specific license in a plugin

ponsfrilus:
The doc says: 
```
If a plugin or theme specifies a license in their main plugin file or stylesheet, the comment looks like
this: 
Copyright (C) 2018 Example Plugin Author
This file is distributed under the GPLv2.
```
but it dosen't seems to work for plugins. It looks like the "License" param in get_file_headers was missing.

### DIFF
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -488,6 +488,7 @@ class MakePotCommand extends WP_CLI_Command {
 					'Author',
 					'Author URI',
 					'Version',
+					'License',
 					'Domain Path',
 					'Text Domain',
 				];


swissspidy:
Thanks for your PR, @ponsfrilus!

Would you be up for providing a Behat test for this change?

ponsfrilus:
@swissspidy I've got some trouble generating a test plugin with a specific license, see [wp-cli/scaffold-command/issues/246](https://github.com/wp-cli/scaffold-command/issues/246).

ponsfrilus:
@swissspidy please see the PR #199 — should be OK now. And sorry for duplicating it, it was not intended.

PR #199, opened 2019-12-12
Ensure that the pot file use the same license as the plugin

ponsfrilus:
It's green :)
image
image

This PR now includes a test case that ensure that the licensing information in the pot file is taken from the plugin file header if exists. Fixes #198.


schlessera:
Thanks for the PR, @ponsfrilus !

PR #204, opened 2020-03-01
fix(make-pot): parses mangled webpack statements (closes #203)

### LINKED ISSUE #203: Webpack mangle is not parsed
## Bug Report

- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

I am using webpack (in my WordPress boilerplate https://github.com/devowlio/wp-react-starter) and want to extract strings (`make-pot`) from my bundled `.js` file. My file contains this code:

```js
Object(_utils__WEBPACK_IMPORTED_MODULE_7__["__"])("Try again"))
```

As long as it looks like this, it works as expected. The coding covers that case:

https://github.com/wp-cli/i18n-command/blob/79d9c75cada495550eb6190e87d523b01b803739/src/JsFunctionsScanner.php#L212-L239

But, due to the fact I need to rely on webpack's `splitChunks` and some other `optimizations` (`usedExports` and `sideEffects`) it no longer works because the output looks like this:

```js
Object(_utils__WEBPACK_IMPORTED_MODULE_7__[/* __ */ "a"])("Try again"))
```

You see, it creates an inline comment `/* __ */` with the function name (can be also `_n` or whatever).

**Describe how other contributors can replicate this bug**

I can not tell you the exact webpack configuration as it is very complicated, but you could do the following:

- I think you have a webpack example project, navigate to it
- Open the bundled `.js` file
- Search for a working `__WEBPACK_IMPORTED_MODULE`
- Replace `"__"` with `/* __ */ "a"`

Edit: You can also create a `test.js` file with the following content:

```js
Object(_utils__WEBPACK_IMPORTED_MODULE_6__[/* __*/ "a"])("Try again");
Object(_utils__WEBPACK_IMPORTED_MODULE_6__["__"])("This works");
```

**Describe what you would expect as the correct outcome**

Should parse the string `Try again` correctly to `.pot` file.

**Let us know what environment you are running this on**

```
OS:     Linux 4.4.0-18362-Microsoft #476-Microsoft Fri Nov 01 16:53:00 PST 2019 x86_64
Shell:  /bin/bash
PHP binary:     /usr/bin/php7.2
PHP version:    7.2.24-0ubuntu0.18.04.2
php.ini used:   /etc/php/7.2/cli/php.ini
WP-CLI root dir:        phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:      phar://wp-cli.phar/vendor
WP_CLI phar path:       /d/Programme/VSCode Workspace/devowl-wp/plugins/devowl-site
WP-CLI packages dir:
WP-CLI global config:
WP-CLI project config:
WP-CLI version: 2.2.0
```

**Provide a possible solution**

I do not exactly know how that works with `Peast` but perhaps we can check for the inline comment (trimmed).

matzeeable:
When working with tree-shaking, vendor chunks and other optimizations webpack will create mangled function names for `__`, `_n`...

```js
Object(_utils__WEBPACK_IMPORTED_MODULE_6__[/* __*/ "a"])("Try again");
```

Reference: #203.

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -1592,6 +1592,7 @@ Feature: Generate a POT file of a WordPress project
       translate.__( 'translate.__', 'foo-plugin' );
 
       Object(_wordpress_i18n__WEBPACK_IMPORTED_MODULE_7__["__"])( 'webpack.__', 'foo-plugin' );
+      Object(_wordpress_i18n__WEBPACK_IMPORTED_MODULE_7__[/* __ */ "a"])( 'webpack.mangle.__', 'foo-plugin' );
 
       Object(u.__)( 'minified.__', 'foo-plugin' );
       Object(j._x)( 'minified._x', 'minified._x_context', 'foo-plugin' );
@@ -1659,6 +1660,10 @@ Feature: Generate a POT file of a WordPress project
     And the foo-plugin/foo-plugin.pot file should contain:
       """
       msgid "webpack.__"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgid "webpack.mangle.__"
       """
     And the foo-plugin/foo-plugin.pot file should contain:
       """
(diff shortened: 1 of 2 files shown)


swissspidy:
Thanks a lot!

Would you also be able to add some tests for this change?

matzeeable:
@swissspidy Tests added and CI passed 😉

Issue #95, opened 2018-10-27
Add option to skip PHP string extraction


swissspidy:
In addition to skipping JS string extraction and skipping string audits (#94) I was thinking about a `--skip-php` option.

This could be useful when one only wants to extract strings from JS files (for whatever reason).

Obviously plugin/theme meta data would still need to be scanned to find the destination folder etc.

n8finch:
I'm working on this for contributor day...

PR #207, opened 2020-04-01
Add option `--skip-php` to `i18n make-pot` command

pmbaldha:
Fix: https://github.com/wp-cli/i18n-command/issues/95


schlessera:
Thanks for the PR, @pmbaldha !

PR #210, opened 2020-04-17
block.json string extraction

### LINKED ISSUE #163: Extract translatable strings from block.json files
## Feature Request

For supporting the next iteration of Gutenberg block development, we need to extract strings from `block.json` files.

`wp i18n make-pot` should parse `block.json` files and extract all strings it finds with the additional context.

For more information, check out:

- https://github.com/WordPress/gutenberg/pull/13693
- https://github.com/WordPress/gutenberg/issues/15169

swissspidy:
Fixes #163.

Todo:

- [x] make tests pass

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -2478,101 +2478,7 @@ Feature: Generate a POT file of a WordPress project
         "icon": "star",
         "description": "Shows warning, error or success notices  ...",
         "keywords": [ "alert", "message" ],
-        "textDomain": "foo-plugin",
-        "attributes": {
-          "message": {
-            "type": "string",
-            "source": "html",
-            "selector": ".message"
-          }
-        },
-        "styleVariations": [
-          { "name": "default", "label": "Default", "isDefault": true },
-          { "name": "other", "label": "Other" }
-        ],
-        "editorScript": "build/editor.js",
-        "script": "build/main.js",
-        "editorStyle": "build/editor.css",
-        "style": "build/style.css"
-      }
-      """
-
-    When I try `wp i18n make-pot foo-plugin`
-    Then STDOUT should be:
-      """
-      Plugin file detected.
-      Success: POT file successfully generated!
-      """
-    And the foo-plugin/foo-plugin.pot file should exist
-    And the foo-plugin/foo-plugin.pot file should contain:
-      """
-      msgid "Foo Plugin"
-      """
-    And the foo-plugin/foo-plugin.pot file should contain:
-      """
-      msgctxt "block title"
-      """
-    And the foo-plugin/foo-plugin.pot file should contain:
-      """
-      msgid "Notice"
-      """
-    And the foo-plugin/foo-plugin.pot file should contain:
-      """
-      msgctxt "block description"
-      """
-    And the foo-plugin/foo-plugin.pot file should contain:
-      """
-      msgid "Shows warning, error or success notices  ..."
-      """
-    And the foo-plugin/foo-plugin.pot file should contain:
-      """
-      msgctxt "block keyword"
-      """
-    And the foo-plugin/foo-plugin.pot file should contain:
-      """
-      msgid "alert"
-      """
-    And the foo-plugin/foo-plugin.pot file should contain:
-      """
-      msgid "message"
-      """
-    And the foo-plugin/foo-plugin.pot file should contain:
-      """
-      msgctxt "block style variation"
-      """
-    And the foo-plugin/foo-plugin.pot file should contain:
-      """
-      msgid "Default"
-      """
-    And the foo-plugin/foo-plugin.pot file should contain:
-      """
-      msgctxt "block style variation"
-      """
-    And the foo-plugin/foo-plugin.pot file should contain:
-      """
-      msgid "Other"
-      """
-
-  Scenario: Extract strings from block.json files with styleVariations alias
-    Given an empty foo-plugin directory
-    And a foo-plugin/foo-plugin.php file:
-      """
-      <?php
-      /**
-       * Plugin Name: Foo Plugin
-       */
-      """
-    And a foo-plugin/block.json file:
-      """
-      {
-        "name": "my-plugin/notice",
-        "title": "Notice",
-        "category": "common",
-        "parent": [ "core/group" ],
-        "icon": "star",
-        "description": "Shows warning, error or success notices  ...",
-        "keywords": [ "alert", "message" ],
-        "textDomain": "foo-plugin",
+        "textdomain": "foo-plugin",
         "attributes": {
           "message": {
             "type": "string",
@@ -2632,7 +2538,7 @@ Feature: Generate a POT file of a WordPress project
       """
     And the foo-plugin/foo-plugin.pot file should contain:
       """
-      msgctxt "block style variation"
+      msgctxt "block style"
       """
     And the foo-plugin/foo-plugin.pot file should contain:
       """
@@ -2640,7 +2546,7 @@ Feature: Generate a POT file of a WordPress project
       """
     And the foo-plugin/foo-plugin.pot file should contain:
       """
-      msgctxt "block style variation"
+      msgctxt "block style"
       """
     And the foo-plugin/foo-plugin.pot file should contain:
       """
@@ -2666,7 +2572,7 @@ Feature: Generate a POT file of a WordPress project
         "icon": "star",
         "description": "Shows warning, error or success notices  ...",
         "keywords": [ "alert", "message" ],
-        "textDomain": "my-plugin",
+        "textdomain": "my-plugin",
         "attributes": {
           "message": {
             "type": "string",
@@ -2674,7 +2580,7 @@ Feature: Generate a POT file of a WordPress project
             "selector": ".message"
           }
         },
-        "styleVariations": [
+        "styles": [
           { "name": "default", "label": "Default", "isDefault": true },
           { "name": "other", "label": "Other" }
         ],
@@ -2739,7 +2645,7 @@ Feature: Generate a POT file of a WordPress project
             "selector": ".message"
           }
         },
-        "styleVariations": [
+        "styles": [
           { "name": "default", "label": "Default", "isDefault": true },
           { "name": "other", "label": "Other" }
         ],
@@ -2816,7 +2722,7 @@ Feature: Generate a POT file of a WordPress project
             "selector": ".message"
           }
         },
-        "styleVariations": [
+        "styles": [
           { "name": "default", "label": "Default", "isDefault": true },
           { "name": "other", "label": "Other" }
         ],
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -2458,3 +2458,81 @@ Feature: Generate a POT file of a WordPress project
       Success: POT file successfully generated!
       """
     And the contents of the result.pot file should match /^msgid/
+
+  Scenario: Extract strings from block.json files
+    Given an empty foo-plugin directory
+    And a foo-plugin/foo-plugin.php file:
+      """
+      <?php
+      /**
+       * Plugin Name: Foo Plugin
+       */
+      """
+    And a foo-plugin/block.json file:
+      """
+      {
+        "name": "my-plugin/notice",
+        "title": "Notice",
+        "category": "common",
+        "parent": [ "core/group" ],
+        "icon": "star",
+        "description": "Shows warning, error or success notices  ...",
+        "keywords": [ "alert", "message" ],
+        "textDomain": "my-plugin",
+        "attributes": {
+          "message": {
+            "type": "string",
+            "source": "html",
+            "selector": ".message"
+          }
+        },
+        "styleVariations": [
+          { "name": "default", "label": "Default", "isDefault": true },
+          { "name": "other", "label": "Other" }
+        ],
+        "editorScript": "build/editor.js",
+        "script": "build/main.js",
+        "editorStyle": "build/editor.css",
+        "style": "build/style.css"
+      }
+      """
+
+    When I try `wp i18n make-pot foo-plugin`
+    Then STDOUT should be:
+      """
+      Plugin file detected.
+      Success: POT file successfully generated!
+      """
+    And the foo-plugin/foo-plugin.pot file should exist
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgid "Foo Plugin"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgctxt "block title"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgid "Notice"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgctxt "block description"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgid "Shows warning, error or success notices  ..."
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgctxt "block keywords"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgid "alert"
+      """
+    And the foo-plugin/foo-plugin.pot file should contain:
+      """
+      msgid "message"
+      """
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -2538,7 +2538,7 @@ Feature: Generate a POT file of a WordPress project
       """
     And the foo-plugin/foo-plugin.pot file should contain:
       """
-      msgctxt "block style"
+      msgctxt "block style label"
       """
     And the foo-plugin/foo-plugin.pot file should contain:
       """
@@ -2546,7 +2546,7 @@ Feature: Generate a POT file of a WordPress project
       """
     And the foo-plugin/foo-plugin.pot file should contain:
       """
-      msgctxt "block style"
+      msgctxt "block style label"
       """
     And the foo-plugin/foo-plugin.pot file should contain:
       """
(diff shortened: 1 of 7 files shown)


src/BlockExtractor.php
```diff
@@ -0,0 +1,59 @@
...
+
+final class BlockExtractor extends Extractor implements ExtractorInterface {
+	use IterableCodeExtractor;
+
+	public static $options = [
+		'translatableProperties' => [
+			'title',
+			'description',
+			'keywords',
+			'styles',
```
gziolo, line 18: I think we ended up with:
```suggestion
			'styleVariations',
```

to make it differ from `style` and `editorStyle`.

src/BlockExtractor.php
```diff
@@ -0,0 +1,61 @@
...
+			return;
+		}
+
+		foreach ( $file_data as $key => $original ) {
+			if ( ! array_key_exists( $key, $options['translatableProperties'] ) ) {
+				continue;
+			}
+
+			foreach ( (array) $original as $msg ) {
+				$translation = $translations->insert( sprintf( 'block %s', $key ), $msg );
```
gziolo, line 56: We need to account for the `styleVariations` objects as well:
```suggestion
				if ( is_object( $msg ) ) {
					$translation = $translations->insert( sprintf( 'block %s %s', $key, $msg->name ), $msg->label );
				} else {
					$translation = $translations->insert( sprintf( 'block %s', $key ), $msg );
				}
```

features/makepot.feature
```diff
@@ -2458,3 +2458,225 @@ Feature: Generate a POT file of a WordPress project
...
+      msgid "message"
+      """
+
+  Scenario: Ignores block.json files with other text domain
+    Given an empty foo-plugin directory
+    And a foo-plugin/foo-plugin.php file:
+      """
+      <?php
+      /**
+       * Plugin Name: Foo Plugin
```
gziolo, line 2546: I wasn't aware that the text domain is auto-generated from the name of the plugin.

features/makepot.feature
```diff
@@ -2458,3 +2458,225 @@ Feature: Generate a POT file of a WordPress project
...
+      """
+      msgid "alert"
+      """
+    And the foo-plugin/foo-plugin.pot file should not contain:
+      """
+      msgid "message"
+      """
+
+
+  Scenario: Extract strings from block.json files with no text domain specified
```
gziolo, line 2607: Does it mean that the text domain is inherited from the plugin? Maybe we should skip the `textDomain` in the `block.json` at all if it shouldn't be different than the one used by the plugin?

src/MakePotCommand.php
```diff
@@ -601,6 +609,20 @@ protected function extract_strings() {
...
 				);
 			}
+
+			if ( ! $this->skip_json ) {
+				BlockExtractor::fromDirectory(
+					$this->source,
+					$translations,
+					[
+						// Only look for block.json files, nothing else.
+						'restrictFileNames' => [ 'block.json' ],
```
gziolo, line 619: @aduth, if you still think that we should allow different names here other than `block.json, it's time to raise it now. I would be personally fine with allowing also `*.block.json` but it's it nice to have rather than something we desperately need.

src/BlockExtractor.php
swissspidy: Do you have an example for `styleVariations`?

features/makepot.feature
swissspidy: Not exactly.

By default, the "Text Domain" header of the plugin or theme is used. If none is provided, it falls back to the project slug (the folder name). That might not always be desirable, so one can also specify the domain using `wp make-pot --domain=foo-bar`.

src/MakePotCommand.php
swissspidy: We can always change this later, or even make it configurable.

features/makepot.feature
swissspidy: In this scenario, yes. According to the RFC, the `textDomain` field is optional. This test ensures that the `block.json` file will be scanned if it's missing.

However, if it's _not missing_ and differs from the one specified earlier in the process, it will be skipped. 

Not sure in which circumstances that happens, but there might be valid use cases for that. Just like people sometimes use multiple domains in PHP & JS code.

src/MakePotCommand.php
aduth: I'm not concerned.

src/BlockExtractor.php
gziolo: You have it included in Behat tests added in PR 😃

swissspidy: 🤦 😅

features/makepot.feature
gziolo: Works for me, we will learn how it looks like in practice.

gziolo:
Thanks for working on it @swissspidy. How can I help to make the tests pass? :)

src/MakePotCommand.php
```diff
@@ -198,6 +202,9 @@ class MakePotCommand extends WP_CLI_Command {
 	 * [--skip-php]
 	 * : Skips PHP string extraction.
 	 *
+	 * [--skip-json]
```
schlessera, line 205: Is `--skip-json` the best flag name here? This might be misleading with regards to extracting _into_ a JSON file.

Also, this doesn't control a `JsonExtractor`, but rather a `BlockExtractor`.

I think `--skip-json` should either control extraction from _any_  JSON file, or it should be renamed to what it actually controls, like `--skip-blocks`.

However, `--skip-blocks` would deviate from the "filetype of input file" approach we currently have for this.

swissspidy: Yeah this is a bit inconsistent because I renamed the extractor mid-flight. Maybe `skip-block-json` to make it more explicit?

schlessera: Yes, would be less misleading at least.

swissspidy: Done in b3c9bd711778daef514bd395746ba7f1349e80c1

swissspidy:
@gziolo I first need to get the tests setup properly running again on my machine, then I can debug them.

swissspidy:
@gziolo Tests are passing now, so this should be ready soon 🙂 But I also just learned about https://github.com/WordPress/gutenberg/pull/21490...

gziolo:
> @gziolo Tests are passing now, so this should be ready soon 🙂 But I also just learned about [WordPress/gutenberg#21490](https://github.com/WordPress/gutenberg/pull/21490)...

Oh, interesting. It looks like more use cases pop up 😄 I don't think `theme.json` is stable enough to wait for it.

src/BlockExtractor.php
```diff
@@ -0,0 +1,71 @@
...
+						continue 2;
+					}
+
+					foreach ( $original as $msg ) {
+						$translation = $translations->insert( 'block keyword', $msg );
+						$translation->addReference( $file );
+					}
+
+					break;
+				case 'styleVariations':
```
swissspidy, line 59: Should check for `styles` too...? Per https://github.com/wp-cli/i18n-command/issues/163#issuecomment-571968841

cc @gziolo

src/BlockExtractor.php
```diff
@@ -0,0 +1,71 @@
...
+
+final class BlockExtractor extends Extractor implements ExtractorInterface {
+	use IterableCodeExtractor;
+
+	/**
+	 * @inheritdoc
+	 */
+	public static function fromString( $string, Translations $translations, array $options = [] ) {
+		$file = $options['file'];
+		WP_CLI::debug( "Parsing file {$file}" );
```
schlessera, line 18: Debug statements should include a group as the second argument.

src/BlockExtractor.php
gziolo: Good catch, let's support both to align with https://github.com/WordPress/gutenberg/blob/master/docs/rfc/block-registration.md#style-variations

> Property: `styles`
> Alias: `styleVariations`

src/IterableCodeExtractor.php
```diff
@@ -23,12 +23,20 @@ trait IterableCodeExtractor {
...
 	 *     Optional. An array of options passed down to static::fromString()
 	 *
-	 *     @type bool $wpExtractTemplates Extract 'Template Name' headers in theme files. Default 'false'.
+	 *     @type bool  $wpExtractTemplates Extract 'Template Name' headers in theme files. Default 'false'.
+	 *     @type array $restrictFileNames  Skip all files which are not included in this array.
 	 * }
 	 * @return null
 	 */
 	public static function fromFile( $file, Translations $translations, array $options = [] ) {
 		foreach ( static::getFiles( $file ) as $f ) {
```
schlessera, line 32: Abbreviations like `$f` make the code hard to read. This should have a full name.

The problem seems to be that the proper variable name `$file` is already taken by the other variable, which actually represents "one ore more file or file paths". So either rename the other one so that this one can be called `$file`, or call this one something like `$individual_file`.

src/BlockExtractor.php
swissspidy: We use `make-pot` as the group elsewhere already. Will re-use that onne.

src/IterableCodeExtractor.php
swissspidy: Done in 119fcf053e3e308532641ba43d82a420b4cae162

src/BlockExtractor.php
swissspidy: Done in 0b2e92bf8d1dbdc43286a8b6e81fc826b8355875

src/BlockExtractor.php
swissspidy: Done in 2507e8db3459eb6fb3cd6e88b6ff7b8c3386b8b1

gziolo:
It looks like all feedback was addressed. When can we expect this PR to be officially released?
I'm wondering when we can finish migration to `block.json` for Gutenberg. It was the last missing ingredient data-wise. :)

I don't see any posts about the next planned minor/major release on the blog. The last release was announced in November:
https://make.wordpress.org/cli/2019/11/12/wp-cli-v2-4-0-release-notes/

schlessera:
@gziolo	I'm about to announce the next release. There were many issues that needed to be fixed first with regards to PHP and MySQL compatibility, but I've had complete passing tests yesterday, so I feel confident enough about announcing a release date now.

I'll merge the above and work on getting it into a nightly release for now, so you can already use and integrate it to run tests.

gziolo:
@swissspidy, one more thing. Can you include an example of how we should wrap those translatable fields with i18n functions? This is how I understand it:

_Example:_

```json
{
    "title": "My block title"
}
```

_In JS code:_

```js
import { _x } from '@wordpress/i18n';
import { title } 'block.json';

console.log( _x( title, 'block title', 'my-plugin' ) );
```

I'm still not sure how it's going to be integrated with the `@wordpress/blocks` package to ensure backward compatibility. We can always create a special utility method like `register_block_type_from_metadata` that would automate the application of i18n functions.

schlessera:
> `console.log( _x( title, 'block title', 'my-plugin' ) );`

Hmm, shouldn't these context names be provided via an interface of some sort (a list of constants, an enum, ...)? Ideally, there would be some form of interface provided by WordPress Core/Gutenberg that defines these in such a way they can be referenced via both PHP & JS.

Otherwise, we have just one more way of producing an architecture where typos result in silent misbehavior that is hard to debug.

gziolo:
> Hmm, shouldn't these context names be provided via an interface of some sort (a list of constants, an enum, ...)? Ideally, there would be some form of interface provided by WordPress Core/Gutenberg that defines these in such a way they can be referenced via both PHP & JS.

This is exactly what I tried to express in https://github.com/wp-cli/i18n-command/pull/210#issuecomment-628540213 with:

> I'm still not sure how it's going to be integrated with the `@wordpress/blocks` package to ensure backward compatibility. We can always create a special utility method like `register_block_type_from_metadata` that would automate the application of i18n functions.

I have a working proposal in the Gutenberg plugin for `register_block_type_from_metadata` on PHP side of things:

https://github.com/WordPress/gutenberg/blob/928ac7bc9e8ce9c07ff929546d3033db16324f3c/lib/compat.php#L26-L44

and it could internally do all the mappings to ensure there are no typos and it's never leaked outside.

schlessera:
> it could internally do all the mappings to ensure there are no typos and it's never leaked outside.

Then the definitions would not be available to WP-CLI, though, and still forces us to hard-code them into WP-CLI, making it impossible to keep in perfect sync across changes.

schlessera:
Related: https://github.com/WordPress/gutenberg/pull/21490#issuecomment-628544366

gziolo:
> > it could internally do all the mappings to ensure there are no typos and it's never leaked outside.
> 
> Then the definitions would not be available to WP-CLI, though, and still forces us to hard-code them into WP-CLI, making it impossible to keep in perfect sync across changes.

I think I didn't understand your concern. We probably talk about two different things. I wasn't aware that `theme.json` wants to expose the list of translatable fields. What would be the ideal flow from WP-CLI side that would make it easy to sync? It might turn out that it'd also make the process of applying those translation functions easier in the core itself :)

schlessera:
My concern is basically that, right now, we have a hard-coded set of attributes within `blocks.json` that needs to be extracted, and it uses a hard-coded set of translation context values to do so.

This means that both of these hard-coded sets need to be kept in sync between Core and WP-CLI, and if you add a new attribute to Core, it will not be picked up by WP-CLI until you've downloaded a new release that knows about this new attribute. It will just silently skip new attributes by default.

The solution to that would be to have a schema definition of some sort, so that we don't define both of these sets within both Core and WP-CLI (and need to keep them in sync), but that we'd rather encode via this schema definition _how to identify_ translatable fields within these JSON files, so that if you add a new attribute in Core, it is automatically picked up by WP-CLI, even at an existing version.

It could be a separate schema definition (maybe using something like [JSON Schema](https://json-schema.org/)?), or it could be a convention within the JSON files themselves (like prefixing translatable keys with an underscore).

The main goal: make it so that we don't need code changes in multiple projects when the list of attributes grows.

gziolo:
I think we are on the same page now. I totally agree that it would be the optimal approach. I also see how Core could use the same schema internally to figure out how to modify loaded `block.json` to automatically apply translations based on the contract between Core and WP-CLI. Great idea. I could come up with that earlier. :)

I personally prefer JSON Schema, because using underscores in the `block.json` would be inconvenient for block developers.

gziolo:
There is now REST API endpoint for block types merged in the Gutenberg repository. It should become a part of WordPress core as part of WordPress 5.5 release. Related PR:
https://github.com/WordPress/gutenberg/pull/21065

I'm not an expert, but it looks like `OPTIONS` returns JSON schema for the shape that this endpoint returns. Would it be possible to extend it to include information on which fields are translatable?

swissspidy:
This kind of schema would need to live somewhere outside of WordPress context, where WP-CLI and other tools can easily assess which fields in `block.json` are translatable and which aren't.

----

In any case, I feel like this discussion should be better held in some Gutenberg GH issue or something, and shouldn't necessarily block this PR.

features/makepot.feature
```diff
@@ -2458,3 +2458,391 @@ Feature: Generate a POT file of a WordPress project
...
+      """
+      {
+        "name": "my-plugin/notice",
+        "title": "Notice",
+        "category": "common",
+        "parent": [ "core/group" ],
+        "icon": "star",
+        "description": "Shows warning, error or success notices  ...",
+        "keywords": [ "alert", "message" ],
+        "textDomain": "foo-plugin",
```
gziolo, line 2481: I discovered today that `textdomain` is the preferred way to phrase it in WordPress core, see my related comment: https://core.trac.wordpress.org/ticket/48529#comment:15

Do you think we should leave it as is or it makes sense to update?

features/makepot.feature
```diff
@@ -2458,3 +2458,391 @@ Feature: Generate a POT file of a WordPress project
...
+        "attributes": {
+          "message": {
+            "type": "string",
+            "source": "html",
+            "selector": ".message"
+          }
+        },
+        "styleVariations": [
```
gziolo, line 2489: We also settled on `styles` to avoid aliases and respect the first choice. The reasoning behind it is as follows: even if it is confusing in the first place, it's also double as much confusing with an alias :)

gziolo:
I landed changes in WordPress core https://github.com/WordPress/wordpress-develop/commit/822ca9ebc7409abce791e4318626ea0bea75fe88 that makes the changes I mentioned lately:
- `styles`
- `textdomain`

There is a corresponding PR that aligns Gutenberg code and updates RFC for block registration: https://github.com/WordPress/gutenberg/pull/22695.

swissspidy:
@gziolo Awesome! Updated the code here accordingly.

src/BlockExtractor.php
```diff
@@ -0,0 +1,72 @@
...
+					}
+
+					break;
+				case 'styles':
+					if ( ! is_array( $original ) ) {
+						continue 2;
+					}
+
+					foreach ( $original as $msg ) {
+						$translation = $translations->insert( 'block style style', $msg['label'] );
```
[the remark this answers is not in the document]
swissspidy: Whoops, I actually meant to only use `block style`.

Perhaps `block style label`? As the word "variation" seems to have been removed everywhere else.

gziolo:
> This kind of schema would need to live somewhere outside of WordPress context, where WP-CLI and other tools can easily assess which fields in block.json are translatable and which aren't.

It looks like JSON schema is in works for the purpose of validation in Block Directory:
https://github.com/tellyworth/plugin-directory/pull/2/files#diff-73a27cb25a34ab6bcbb40bbbc9c436afR35

Where would be the best place to keep the common schema? I guess that Block Directory sooner or later would need to add its own flavors :)

swissspidy:
Probably on WordPress.org somewhere

gziolo:
> This kind of schema would need to live somewhere outside of WordPress context, where WP-CLI and other tools can easily assess which fields in `block.json` are translatable and which aren't.

I opened a ticket in WordPress track to continue the discussion: https://core.trac.wordpress.org/ticket/50615.

PR #211, opened 2020-05-04
WIP: Gettext 5

drzraf:
Roadblock of https://github.com/php-gettext/Twig-Scanner
(back from https://github.com/wp-cli/i18n-command/pull/59)

- Should the PR be based on wp-cli:dependabot/composer/gettext/gettext-tw-5.4?
- Any particular point to consider before I further improve that PR?


swissspidy:
I'll let @schlessera chime in, but I think the PHP 7.2+ requirement of gettext v5 might be a bit of a blocker for us here.

schlessera:
Yes, indeed, moving to the v5 branch of `gettext` is currently not an option, as WP-CLI is supporting PHP 5.4+ for now (with an imminent bump to PHP 5.6+ upcoming).

See https://make.wordpress.org/cli/2019/01/15/wp-cli-php-requirements-strategy/ about details on how the minimum requirement is decided.

Feel free to hit me up on Slack if you want to discuss options about how best to enable Twig support despite this limitation.

drzraf:
Ok.
Which version of PHP 7.x will be required in v3? 7.1, 7.2, ... ?
Is work on v3 going to be done inside v3 branches for all packages?

If yes, and >= PHP 7.2, then I'd like to anticipate that move (so that I can just require `i18n-command:^3` in my `composer.json`)

drzraf:
@dependabot-preview /reopen please.

drzraf:
Still supporting PHP < 7.2 ?

PR #217, opened 2020-06-18
WIP: gettext/gettext 5.x

drzraf:
Roadblock of https://github.com/php-gettext/Twig-Scanner
(back from https://github.com/wp-cli/i18n-command/pull/59 and #211)

Note:
- Gettext 5 is PHP 7.2

Questions:
- Which version of PHP 7.x will be required in v3 in December? 7.1, 7.2, ... ?
- Can a gettext-5 already exists somewhere in this repository? Or within a wp-cli-v3 branch?


schlessera:
> Which version of PHP 7.x will be required in v3 in December? 7.1, 7.2, ... ?

Current policy is to stick to WordPress minimum with a lag of 1 year. So as WordPress is currently still at the 5.6 minimum, WP-CLI will certainly be at 5.6 still in December.

> Can a gettext-5 already exists somewhere in this repository? Or within a wp-cli-v3 branch?

Yes, we could create a separate branch in `wp-cli/i18-command` for that. You should be able to install it via the package manager then, I think.

drzraf:
Since June 2020, the minimum "acceptable_php" version is 7.2
(https://meta.trac.wordpress.org/ticket/5257)
Could we anticipate that by June 2021, you would accept a PR depending upon gettext 5?

schlessera:
What matters here is not the `acceptable_php` version (which is the one that decides whether users get a nag screen they should update their PHP version), but rather the actually **minimum** PHP version. Any code that requires more than the **minimum** PHP version will break for a portion of the users.

We can make this change 1 year from the point where WordPress has bumped its minimum PHP version to 7.2, I'm afraid. Anything else would lock people out of WP-CLI, which is a critical administrative tool first of all.

I understand your frustration, and I share it. WordPress was meant to bump the PHP minimum to 7+ in December of last year. But that hasn't happened so far... I'm having a lot of trouble keeping everything working across all of the currently supported versions, and I'd love getting rid of PHP 5.

drzraf:
Would a v2 / PHP-7.x branch of i18n be an option?

swissspidy:
> Would a v2 / PHP-7.x branch of the `i18n` command be an option?

Meaning we'd have to maintain two versions for quite some time?

schlessera:
@drzraf This PR was moved to the `gettext-5` branch within the WP-CLI repo. This way, you should be able to pull it in via Composer. Feel free to create PRs against that `gettext-5` branch as needed, I'll merge them in quick turn-arounds, as this is not directly impacting WP-CLI compatibility.

PR #222, opened 2020-07-24
Extract strings contained within eval()

dd32:
Some build systems can result in the built files gettext functions being inside eval() calls.

One such example of this, is a webpack'd built file in this plugin: https://plugins.trac.wordpress.org/browser/mathml-block/trunk/dist/mathml-block.js?rev=2276811#L245

Currently, the relevant output is this:
```
#: src/mathml-block.js:35
msgid "MathML formula:"
msgstr ""

#: src/mathml-inline.js:10
msgid "MathML"
msgstr ""
```

which doesn't work for WordPress Translation files, as those src files are never enqueued, only the built files.
After the change, here's the output:
```
#: dist/mathml-block.js:245
#: src/mathml-block.js:35
msgid "MathML formula:"
msgstr ""

#: dist/mathml-block.js:257
#: src/mathml-inline.js:10
msgid "MathML"
msgstr ""
```

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -1616,7 +1616,7 @@ Feature: Generate a POT file of a WordPress project
       Object(u.__)( 'minified.__', 'foo-plugin' );
       Object(j._x)( 'minified._x', 'minified._x_context', 'foo-plugin' );
 
-      eval("__( 'Hello Eval World', 'foo-plugin' );" );
+      eval( "__( 'Hello Eval World', 'foo-plugin' );" );
       """
 
     When I run `wp i18n make-pot foo-plugin`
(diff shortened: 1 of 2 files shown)


dd32:
The unit tests keep timing out on trunk only - I'm not sure why as I've not been able to run then locally.

swissspidy:
Interesting edge case & nice solution for the line number!

Would you be able to add some Behat tests for this? Could probably just copy some lines from the MathML plugin and then verify that strings are extracted as expected with the correct line number.

dd32:
I'll get some tests added @swissspidy - Hopefully I can figure out how to run the tests locally

dd32:
Unfortunately I wasn't able to get tests running locally, and there are no existing tests for the JS set of extraction methods, so I can't really fly into it blind either.

swissspidy:
There are no unit tests for it, but Behat tests, starting here:

https://github.com/wp-cli/i18n-command/blob/b02ecdc9a57f9633740c254d19749118b7411f0f/features/makepot.feature#L1559-L1580

If it's not working, feel free to add me to your fork so I could amend, otherwise I can open a new PR with your commit in it and the tests.

swissspidy (approved):

swissspidy:
@schlessera Any idea about the test timeouts? I can't re-start Travis in case it's just a flake.

PR #249, opened 2021-05-25
Add support to IterableCodeExtractor for symlinks.

jenkoian:
Could only get the tests to pass locally with the [yoast polyfill TestCase](https://github.com/Yoast/PHPUnit-Polyfills#option-1-yoastphpunitpolyfillstestcasestestcase) but let me know if you want me to revert to `PHPUnit_Framework_TestCase`

### DIFF
--- a/tests/IterableCodeExtractorTest.php
+++ b/tests/IterableCodeExtractorTest.php
@@ -2,11 +2,11 @@
 
 namespace WP_CLI\I18n\Tests;
 
-use PHPUnit_Framework_TestCase;
 use WP_CLI\I18n\IterableCodeExtractor;
+use WP_CLI\Tests\TestCase;
 use WP_CLI\Utils;
 
-class IterableCodeExtractorTest extends PHPUnit_Framework_TestCase {
+class IterableCodeExtractorTest extends TestCase {
 
 	/** @var string A path files are located */
 	private static $base;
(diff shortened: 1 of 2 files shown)


swissspidy:
Does this also apply to JS files?

@schlessera Looks like the PHPUnit polyfills are now part of wp-cli-tests, so seems reasonable to use it here. WDYT?

jenkoian:
> Does this also apply to JS files?

Not sure I follow?

tests/IterableCodeExtractorTest.php
```diff
@@ -2,16 +2,16 @@
 
 namespace WP_CLI\I18n\Tests;
 
-use PHPUnit_Framework_TestCase;
 use WP_CLI\I18n\IterableCodeExtractor;
 use WP_CLI\Utils;
+use Yoast\PHPUnitPolyfills\TestCases\TestCase;
 
-class IterableCodeExtractorTest extends PHPUnit_Framework_TestCase {
```
swissspidy, line 9: @schlessera I'm getting `PHP Fatal error:  Uncaught Error: Class 'PHPUnit_Framework_TestCase' not found` on `master` without this change. Any advice?

tests/IterableCodeExtractorTest.php
```diff
@@ -2,16 +2,16 @@
...
+class IterableCodeExtractorTest extends TestCase {
 
 	/** @var string A path files are located */
 	private static $base;
 
-	public function setUp() {
+	public function set_up() {
```
swissspidy, line 14: @jenkoian Can you undo the changes here? The tests will fail locally but pass on CI.

I think this needs to be fixed in a separate PR.

tests/IterableCodeExtractorTest.php
schlessera: Use `WP_CLI\Tests\TestCase` instead, to avoid direct coupling to the polyfills.

tests/IterableCodeExtractorTest.php
jenkoian: Have used the `WP_CLI\Tests\TestCase` as mentioned above and changed this back as you suggested. Was that all?

swissspidy (approved):

Issue #224, opened 2020-09-03
Extract translatable strings from theme.json files


jorgefilipecosta:
## Feature Request

 

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

For supporting the next generation of themes, we need to extract strings from theme.json files.

**Describe the solution you'd like**

wp i18n make-pot should parse theme.json files and extract all strings it finds with the additional context.

Sample file:
```
{
	"global": {
		"presets": {
			"color": [
				{
					"slug": "blue",
					"value": "#00f",
					"name": "Red"
				}
			],
			"font-size": [
				{
					"slug": "small",
					"value": 13,
					"name": "Small"
				}
			],
			"font-family": [
				{
					"value": "-apple-system,BlinkMacSystemFont,\"Segoe UI\",Roboto,Oxygen-Sans,Ubuntu,Cantarell,\"Helvetica Neue\",sans-serif",
					"slug": "system",
					"name": "System"
				}
			],
			"gradient": [
				{
					"slug": "blush-bordeaux",
					"value": "linear-gradient(135deg,rgb(254,205,165) 0%,rgb(254,45,45) 50%,rgb(107,0,62) 100%)",
					"name": "Another gradient"
				}
			]
		}
	},
	"core/paragraph": {
		"presets": {
			"color": [
				{
					"slug": "red",
					"value": "#f00",
					"name": "Red"
				}
			],
			"font-size": [
				{
					"slug": "small",
					"value": 13,
					"name": "Small"
				}
			],
			"font-family": [
				{
					"value": "-apple-system,BlinkMacSystemFont,\"Segoe UI\",Roboto,Oxygen-Sans,Ubuntu,Cantarell,\"Helvetica Neue\",sans-serif",
					"slug": "system",
					"name": "System"
				}
			],
			"gradient": [
				{
					"slug": "blush-bordeaux",
					"value": "linear-gradient(135deg,rgb(254,205,165) 0%,rgb(254,45,45) 50%,rgb(107,0,62) 100%)",
					"name": "Simple gradient"
				}
			]
		}
	}
}
```

In all these cases "name" should be translatable.
For now, these files are named experimental-theme.json (future theme.json) on WordPress themes, and experimental-default-theme.json (future default-theme.json) for the core own default file.

It is very related to the work done for block.json https://github.com/wp-cli/i18n-command/pull/210.

[...]

PR #254, opened 2021-06-21
Extract strings for translation from theme.json

nosolosw:
Closes https://github.com/wp-cli/i18n-command/issues/224

## Setup

- See https://make.wordpress.org/cli/handbook/contributions/pull-requests/#setting-up
- To run the behat tests you also need a working mysql locally:
  - Install the packages if you don't have them. In linux: `sudo apt install jq mysql-server mysql-client php-mysql`
  - `composer install && composer behat` => it didn't setup the database for me (mysql 8) so I did;
    - `sudo mysql -u root`
    - `CREATE DATABASE IF NOT EXISTS `wp_cli_test`;`
    - `CREATE USER 'wp_cli_test'@'localhost' IDENTIFIED WITH mysql_native_password BY 'password1';`
    - `GRANT ALL PRIVILEGES ON wp_cli_test.* TO "wp_cli_test"@"localhost"`

## How to test

- `composer install`
- Create a new directory called `foo-theme` and create a `theme.json` file within with the following contents:

```json
      {
        "version": "1",
        "settings": {
          "color": {
            "palette": [
              { "slug": "black", "color": "#000000", "name": "Black" }
            ]
          }
        }
      }
```

- Verify that there's a `foo-theme/foo-theme.pot` file and that the "Black" string is present as well as its context "Color name". You should see something like:

```
msgctxt "Color name"
msgid "Black"
msgstr ""
```

## Stress test the network connection

Switch to offline and verify that the file is still generated properly using the local file. Upon executing the command, you should see the following message:

```
PHP Warning:  file_get_contents(): php_network_getaddresses: getaddrinfo failed: Temporary failure in name resolution in /home/andres/src/i18n-command/src/ThemeJsonExtractor.php on line 106
PHP Warning:  file_get_contents(https://raw.githubusercontent.com/WordPress/gutenberg/wp/trunk/lib/experimental-i18n-theme.json): failed to open stream: php_network_getaddresses: getaddrinfo failed: Temporary failure in name resolution in /home/andres/src/i18n-command/src/ThemeJsonExtractor.php on line 106
Success: POT file successfully generated!
```


src/theme-i18n.json
```diff
@@ -0,0 +1,51 @@
+{
```
nosolosw, line 1: Ideally, wp-cli would take this file from the WordPress core if it exists. If it doesn't, it shouldn't lookup for strings in the directory, as it means that the WordPress version in use doesn't support `theme.json`.

src/ThemeJsonExtractor.php
```diff
@@ -0,0 +1,175 @@
...
+	 *     _wp_array_get( $array, array( 'a', 'b', 'c' ) );
+	 *
+	 * @param array $array   An array from which we want to retrieve some information.
+	 * @param array $path    An array of keys describing the path with which to retrieve information.
+	 * @param mixed $default The return value if the path does not exist within the array,
+	 *                       or if `$array` or `$path` are not arrays.
+	 *
+	 * @return mixed The value from the path specified.
+	 */
+	private function _wp_array_get( $array, $path, $default = null ) {
```
nosolosw, line 155: This is an utility that I've ported from WordPress core, so wp-cli doesn't depend on it being present.

src/ThemeJsonExtractor.php
```diff
@@ -0,0 +1,175 @@
...
+			}
+
+			if ( is_array( $decoded_file ) ) {
+				$config = $decoded_file;
+			}
+		}
+		return $config;
+	}
+
+	private static function get_fields_to_translate() {
```
nosolosw, line 101: This methods as well as `extract_paths_to_translate` and `read_json_file` have been ported from `WP_Theme_JSON_Resolver` in WordPress core.

src/ThemeJsonExtractor.php
```diff
@@ -0,0 +1,183 @@
...
+	 *     array_get( $array, array( 'a', 'b', 'c' ) );
+	 *
+	 * @param array $array   An array from which we want to retrieve some information.
+	 * @param array $path    An array of keys describing the path with which to retrieve information.
+	 * @param mixed $default The return value if the path does not exist within the array,
+	 *                       or if `$array` or `$path` are not arrays.
+	 *
+	 * @return mixed The value from the path specified.
+	 */
+	private function array_get( $array, $path, $default = null ) {
```
jorgefilipecosta, line 163: Should we link in comments to the WordPress core version of these functions?

src/ThemeJsonExtractor.php
```diff
@@ -0,0 +1,183 @@
...
+use Gettext\Translations;
+use WP_CLI;
+
+final class ThemeJsonExtractor extends Extractor implements ExtractorInterface {
+	use IterableCodeExtractor;
+
+	/**
+	 * @inheritdoc
+	 */
+	public static function fromString( $string, Translations $translations, array $options = [] ) {
```
jorgefilipecosta, line 16: This function for the block.json extractor has a check to skip translation if the translation domains don't match.
https://github.com/dlind1/i18n-command/blob/d4a00c4e97007ffd8233b4ead7e869cf5c4226e1/src/BlockExtractor.php

Should we have a similar check? We don't deal with domains in this class, given that we are extracting strings from a theme are domains automatically handled?

swissspidy: `theme.json` does not have a `domain` field, correct? In that case, the answer would be no.

src/theme-i18n.json
swissspidy: Note that `wp i18n make-pot` is usually run against a single theme's directory, with no knowledge of a WP installation around it.

Ideally the `theme-i18n.json` file is requested from an external API, e.g. WordPress.org, based on the `version` field in the `theme.json` file.

It shouldn't be bundled with WP-CLI for sure.

BTW, the same would need to be done for `block.json`.

src/ThemeJsonExtractor.php
swissspidy, line 163: We might want to move those utils to https://github.com/wp-cli/wp-cli or at least extract them to a new abstract class so it can be used by `BlockExtractor`

src/ThemeJsonExtractor.php
nosolosw: My understanding is that:

1. wp-cli already handles setting up the domain for the translation based on the type of input (in the case of themes, from the stylesheet).
2. while `block.json` provides an additional domain, `theme.json` doesn't, so we can't compare the `theme.json`'s domain to the one inferred by wp-cli.

src/theme-i18n.json
nosolosw: As per https://github.com/wp-cli/i18n-command/issues/224#issuecomment-736646832 by @schlessera my understanding is that we want to access this from WordPress core. What would be the preferred way to do so?

src/ThemeJsonExtractor.php
nosolosw: Happy to create a class for this to be available in other places. I don't see any potential use of them in `BlockExtractor`, though, so perhaps we can extract them when we need them?

nosolosw: done at https://github.com/wp-cli/i18n-command/pull/254/commits/a359599d09db76bc487e6881db0440fda08ba3dd

src/ThemeJsonExtractor.php
nosolosw: What Pascal said. I didn't see his comment until now (apparently, comments are only updated upon page reload).

src/theme-i18n.json
nosolosw: (same as below, didn't see Pascal's comment when posting mine because I hadn't reloaded the browser)

nosolosw: It may take a bit to create and deploy an endpoint in wordpress.org. Would you think it's reasonable to bundle this file with 2.5.1 so we then can update translate.wordpress.org for these strings to be translatable for the WordPress 5.8 release? We can work on that endpoint later, with more time.

nosolosw: > BTW, the same would need to be done for block.json.

I don't know what this means. Is the idea to change the implementation of the existing `BlockExtractor.php` so it takes its data from a similar file?

swissspidy: > I don't know what this means. Is the idea to change the implementation of the existing BlockExtractor.php so it takes its data from a similar file?

Yes. This file already exists in Gutenberg: https://github.com/WordPress/gutenberg/blob/77bc9cae4d016f52523f0ee7018183961ab6056e/packages/blocks/src/api/i18n-block.json

nosolosw: I want to share an alternative proposal: what if we have a PHP package that bundles this logic (code + `theme-i18n.json`) and can be reused across wp-cli, WordPress core, and Gutenberg?

It'd have a few advantages: 1) no need to be online to create a `.pot` file, run WordPress or the plugin; 2) modifications to the `theme-i18n.json` structure that affect the code can be distributed from a single point instead of having to coordinate updates to three codebases.

This would take some time to figure out, so I'd like to propose that we bundle `theme-i18n.json` with this command for the sake of releasing 2.5.1.

nosolosw:
@swissspidy @schlessera do you think this is ready? The only bit under discussion is whether the `theme-18n.json` can be bundled with this command until we find a better path forward.

swissspidy (approved):

nosolosw:
@swissspidy @schlessera what would be the next steps for this? Anything I can do to move things forward?

swissspidy:
I'd love to get some additional feedback from @schlessera 

Once merged, it would be good to follow up with another PR to improve the `block.json` extractor in similar fashion, re-using some of the code.

src/theme-i18n.json
schlessera: Ideally, this file would be fetched online on-demand, with a committed version in this repo being the fallback if no online access or the download failed (which would also show a warning in that case).
The committed fallback here could be kept in sync via a GitHub action, so that we're eventually up-to-date without manual intervention.

schlessera: What we'd need is a reliable URL where this file can be fetched in its latest version. It doesn't necessarily have to be an API, but it should be as stable as possible.

nosolosw: In https://github.com/wp-cli/i18n-command/pull/254/commits/6d850a120f7e6a24fb71a5e7091228497befde32 I'm pulling the file from Gutenberg's GitHub repo. I've also added a fallback to the local file if there was an issue and it is empty. I've tried using WordPress.org SVN but it reports a 403.

If this unblocks us to ship this, let's do it. My preference, however, would be to start only with the local file as it less fragile.

nosolosw: Added some stress testing instructions in the description.

src/ThemeJsonExtractor.php
```diff
@@ -140,7 +142,20 @@ private static function read_json_file( $file_path ) {
...
+		$context        = stream_context_create( array(
+			'http' => array(
+				'method'        => 'GET',
+				'header'        => 'Content-type: application/json',
+				'timeout'       => '3', // To make sure it resolves in a reasonable timeframe.
+			)
+		));
+		// Using the WordPress.org SVN repo resolved to a 403.
+		// $file_structure = self::read_json_file( 'http://develop.svn.wordpress.org/trunk/src/wp-includes/theme-i18n.json', $context );
+		$file_structure = self::read_json_file( 'https://raw.githubusercontent.com/WordPress/gutenberg/wp/trunk/lib/experimental-i18n-theme.json', $context );
```
swissspidy, line 154: `https://raw.githubusercontent.com/WordPress/wordpress-develop/master/src/wp-includes/theme-i18n.json`?

nosolosw: I've used the one in the `wp/trunk` branch because that's the one that only changes with WordPress core. I wish I could have used `wp/5.8` but that branch is not created until WordPress 5.8 is released like WordPress.org's SVN.

swissspidy: So `https://raw.githubusercontent.com/WordPress/wordpress-develop/5.8/src/wp-includes/theme-i18n.json` perhaps?

schlessera: Locking it to 5.8 doesn't help at all, then it can just as well be statically committed here. The idea of an outside source is about getting the changes from Core without needing to release another WP-CLI version in sync. If you lock the version that WP-CLI looks for, there's not much of a point.

schlessera: It should be either:
1. always retrieve the translation info for the latest development version: or
2. always retrieve the translation info for the latest stable version.

Any other version is basically the same as having a hard-coded version inside of WP-CLI.

nosolosw: I can change it to `master` in wordpress-develop.

However, as I shared above, making wp-cli dependant on an external file is fragile: there are changes to the `theme-i18n.json` that would work with the current code but there are other changes that won't.

nosolosw: Done at https://github.com/wp-cli/i18n-command/pull/254/commits/eac73a112a7740b8bde17d82398041375d6dfd51

schlessera: Can you share more details about what type of changes wouldn't work and why can't we make them work?

nosolosw: Any structural change to `theme.json` can affect this. A few ones I can think of right now: translatable strings need to be part of a leaf node that is an unkeyed array, the format supports variability in node names encoded via `*` but only to one level, can't translate several strings from the same leaf, etc.

It's not that these are planned changes. My point is that, in the current state of things, the unit that best encapsulates the behavior is the file + the code used to parse it. Hence, pulling the file from elsewhere is risky: can work with some changes, and won't work with others.

nosolosw: hi @schlessera is there anything I can help with to move this forward?

schlessera: Not at the moment. I'll try to find the time to do a thorough final review, and then it will be good to be merged.
What I'll be looking at most of all, is ways in which this can break if the format changes, and ensure we have graceful handling for such cases (instead of causing a fatal or - even worse - succeeding with the wrong output).

nosolosw: Feel free to push changes to this PR if you want to. I'm also happy to add more graceful handling if we come up with anything.

nosolosw:
@schlessera @swissspidy I wanted to share the last fortnight of July (15th onwards) I'll have very limited availability. JFYI so you know what to expect from me in terms of time to respond/help/feedback/etc. You are very welcome to own this PR, should I not be around.

youknowriad:
Hi folks, what's the status for this one, do you think we're still on track to land this in 5.8?

src/ThemeJsonExtractor.php
```diff
@@ -0,0 +1,268 @@
...
+		$context = stream_context_create(
+			array(
+				'http' => array(
+					'method'  => 'GET',
+					'header'  => 'Content-type: application/json',
+					'timeout' => '3', // To make sure it resolves in a reasonable timeframe.
+				),
+			)
+		);
+		// Using the WordPress.org SVN repo resolved to a 403.
```
ocean90, line 154: Adding a User-Agent header should fix the 403.

swissspidy:
> Hi folks, what's the status for this one, do you think we're still on track to land this in 5.8?

Just awaiting final review from @schlessera

src/theme-i18n.json
ocean90: I agree with @nosolosw that for the first version a bundled schema file should only be used to not depend on an unstable URL like it's implemented right now. It's unstable because there's no version at all. In case there's a change to the structure of the file or the file is moved we'd have to update WP-CLI anyway right?
I don't think it's responsible to block/delay the command from working in this case just because a not yet published update isn't supported.

A compromise might be to add a `--prefer-local-schema` argument which can be used by environments which require a stable behaviour?

src/ThemeJsonExtractor.php
schlessera: I switched the remote get to using `Utils\http_request`, which in turn uses the `Requests` library. SVN works with that.

schlessera (approved):

src/theme-i18n.json
ocean90: @schlessera Any thoughts on a `--prefer-local-schema` argument?

schlessera: @ocean90 Sounds useful, especially if it accepts a file path (which would default to the included one). This would let you easily experiment with different schema.

PR #265, opened 2021-07-05
Add new `--update-mo-files` flag to `make-json`

### LINKED ISSUE #126: Update MO files when purging
When running `wp i18n make-json` with the `--purge` options, we currently modify the PO files without also updating the MO files (if they exist) afterwards.

Perhaps we should do this. Otherwise you have to do this manually for all updated PO files.

swissspidy:
Fixes #126

### DIFF
--- a/src/MakeJsonCommand.php
+++ b/src/MakeJsonCommand.php
@@ -41,6 +41,10 @@ class MakeJsonCommand extends WP_CLI_Command {
 	 * [--purge]
 	 * : Whether to purge the strings that were extracted from the original source file. Defaults to true, use `--no-purge` to skip the removal.
 	 *
+	 * [--update-mo-files]
+	 * : Whether MO files should be updated as well after updating PO files.
+	 * Only has an effect when used in combination with `--purge`.
+	 *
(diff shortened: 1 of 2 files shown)


src/MakeJsonCommand.php
```diff
@@ -102,6 +107,18 @@ public function __invoke( $args, $assoc_args ) {
...
+						continue;
+					}
+
+					if ( $update_mo_files ) {
+						$file_basename    = basename( $file->getFilename(), '.po' );
+						$destination_file = "{$destination}/{$file_basename}.mo";
+
+						$translations = Translations::fromPoFile( $file->getPathname() );
+						if ( ! $translations->toMoFile( $destination_file ) ) {
+							WP_CLI::warning( sprintf( 'Could not create file %s', $destination_file ) );
```
schlessera, line 119:
```suggestion
							WP_CLI::warning( "Could not create file {$destination_file}" );
```

schlessera (changes requested):

schlessera (approved):

PR #267, opened 2021-07-12
make-pot: add `subtract-and-merge` flag

### LINKED ISSUE #105: Try to merge code references when using subtract
Not sure how and if even possible, but would like to raise this request anyway.

The issue was [reported 6 years ago](https://core.trac.wordpress.org/ticket/20881) for makepot.php and is still valid for `i18n make-pot`.

**Example**
`wordpress.pot` includes the string "Edit" like this:

```
#: wp-includes/class-wp-editor.php:1262
#: wp-includes/class-walker-comment.php:236
#: wp-includes/class-walker-comment.php:340
msgid "Edit"
msgstr ""
```

Without subtracting `wordpress.pot`, `wordpress-admin.pot` also includes the string "Edit":

```
#: wp-admin/includes/class-wp-terms-list-table.php:437
#: wp-admin/includes/class-wp-users-list-table.php:410
#: wp-admin/includes/dashboard.php:634
#: wp-admin/includes/post.php:1356
#: wp-admin/includes/class-wp-ms-sites-list-table.php:513
#: wp-admin/includes/class-walker-nav-menu-edit.php:148
#: wp-admin/includes/meta-boxes.php:107
#: wp-admin/includes/meta-boxes.php:154
#: wp-admin/includes/meta-boxes.php:218
#: wp-admin/includes/class-wp-links-list-table.php:317
#: wp-admin/includes/class-wp-ms-users-list-table.php:335
#: wp-admin/includes/class-wp-ms-users-list-table.php:439
#: wp-admin/includes/class-wp-posts-list-table.php:396
#: wp-admin/includes/class-wp-posts-list-table.php:1238
#: wp-admin/includes/class-wp-comments-list-table.php:589
#: wp-admin/includes/class-wp-media-list-table.php:652
#: wp-admin/includes/class-wp-media-list-table.php:701
#: wp-admin/edit-form-comment.php:106
#: wp-admin/comment.php:216
msgid "Edit"
msgstr ""
```

Now when using `--substract=wordpress.pot`, `wordpress-admin.pot` would not contain the "Edit" string but `wordpress.pot` wasn't updated to include the "merged" code references.
I'm aware that it's not a merge so updating the files provided via substract might not something that someone would expect? Maybe do this only if another flag `--substract-and-merge` is provided?

swissspidy:
This PR adds a new `--subtract-and-merge` flag to the `make-pot` command, which can be used in combination with `--subtract`. When used, source code references from the generated POT file will instead be added to the POT file(s) passed to `--subtract`

Example:

```bash
# In foo-plugin/
wp i18n make-pot . foo-plugin.pot --subtract=exceptions.pot

# foo-plugin.pot will not contain any of the strings that are also in exceptions.pot
# exceptions.pot will now however contain all source code references for these strings
```

While I suggested adding dedicated commands for this feature a few years ago, in hindsight thats seems a bit overkill. It's quite the advanced configuration option that's probably only ever used for WordPress core, so just adding a flag feels like the better choice here.

Fixes #105

### DIFF
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -674,9 +674,7 @@ class MakePotCommand extends WP_CLI_Command {
 
 				if ( $this->subtract_and_merge ) {
 					$translation = $translations[ $exception_translation->getId() ];
-					foreach ( $translation->getReferences() as $reference ) {
-						$exception_translation->addReference( $reference[0], $reference[1] );
-					}
+					$exception_translation->mergeWith( $translation );
 				}
 
 				unset( $translations[ $exception_translation->getId() ] );
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -196,7 +196,7 @@ class MakePotCommand extends WP_CLI_Command {
 	 * different content and no duplicate strings between them.
 	 *
 	 * [--subtract-and-merge]
-	 * : Whether source code references from the generated POT file should be instead added to the POT file
+	 * : Whether source code references and comments from the generated POT file should be instead added to the POT file
 	 * used for subtraction. Warning: this modifies the files passed to `--subtract`!
 	 *
 	 * [--include=<paths>]
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -30,9 +30,14 @@ class MakePotCommand extends WP_CLI_Command {
 	protected $merge = [];
 
 	/**
-	 * @var Translations
+	 * @var Translations[]
 	 */
-	protected $exceptions;
+	protected $exceptions = [];
+
+	/**
+	 * @var bool
+	 */
+	protected $subtract_and_merge;
 
 	/**
 	 * @var array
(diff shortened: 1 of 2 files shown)


swissspidy:
@ocean90 Would this solve the issue for dotorg?

ocean90:
> @ocean90 Would this solve the issue for dotorg?

As of https://meta.trac.wordpress.org/changeset/10063/ this is actually no issue anymore. My approach is also preserving any comments by using the `mergeWith()` method. Maybe this should be considered here as well?

swissspidy:
Good suggestion with `mergeWith()` 👍 

Are there chances that you would adopt this new `--subtract-and-merge` flag should it be add? If not, we might not want to bother implementing it.

ocean90:
If if works the same as the current implementation, sure why not. Less code I have to maintain. 😅

swissspidy:
It should work the same way now and is ready for review.

src/MakePotCommand.php
```diff
@@ -190,6 +195,10 @@ class MakePotCommand extends WP_CLI_Command {
 	 * This can be useful when you want to create multiple POT files from the same source directory with slightly
 	 * different content and no duplicate strings between them.
 	 *
+	 * [--subtract-and-merge]
+	 * : Whether source code references from the generated POT file should be instead added to the POT file
```
ocean90, line 199:
```suggestion
	 * : Whether source code references and comments from the generated POT file should be instead added to the POT file
```

ocean90 (changes requested):

schlessera (approved):

PR #270, opened 2021-07-17
fix: Dependency gettext 4.8.5 mixes objects and strings (resolves #269)

### LINKED ISSUE #269: Breaking change in gettext 4.8.5
In v4.8.5 of gettext, [this change](https://github.com/php-gettext/Gettext/pull/271) was made to cache redundant translator comments, but this breaks [`MakePotCommand.php`](https://github.com/wp-cli/i18n-command/blob/master/src/MakePotCommand.php#L683) because `$comment` can now be an object and not a string, if a `translators:` comment is found.

Not sure if the maintainers of this project wanted to reach out to gettext in case this should be backed out in their patch releases, or if the logic here should be changed, or if you'd want to just set `4.8.4` instead of `^4.8` in i18n's `composer.json`

theimbender:
In gettext v4.8.5 it is now possible for the `$comment` variable to be an object here instead of just a string.

### DIFF
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -680,6 +680,9 @@ class MakePotCommand extends WP_CLI_Command {
 						/** @var string $comment */
 						/** @var string $file_header */
 						foreach ( $this->get_file_headers( $this->project_type ) as $file_header ) {
+							if ( is_object( $comment ) && method_exists( $comment, 'getComment' ) ) {
+								$comment = $comment->getComment();
+							}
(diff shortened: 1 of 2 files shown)


swissspidy:
This is a duplicate of #286, but thanks for opening this PR & the issue!

Since I added this change to gettext I might need to look into submitting another PR there.

schlessera:
Thanks for the PR, @theimbender! I went with merging #286 as it is more precise about the typing, but I'll make sure we add you to the list of contributors anyway.

Issue #282, opened 2021-08-27
Suggestion on handling build steps in `make-json`


cpiber:
Currently I believe the accepted way of translating `js` files to run the build steps, then run `make-json` on the built files. As #127 (among others) mentions this loses all context.

As a joint effort with `@wordpress/scripts`, a webpack plugin could be used to solve this problem.

My suggestion is as follows:

- The webpack plugin should emit a `json` file next to the built `js` file containing an object mapping from source file to build file (relative paths like in the `po[t]` file).
- `make-json` would take a new optional argument. This argument would change the behaviour of `make-json` to only extract the files that are keys in the `json` file, and to bundle them by their value (to support multiple build files in one mapper `json`).

This of course requires reading files in before checking `po` files. For flexibility on where the `json` file(s) are, either the new argument could optionally take a glob value, or the `include`/`exclude` arguments could be reused.

PR #284, opened 2021-09-02
make-json Allow to keep source file context by translating to build files during extraction

### LINKED ISSUE #127: Improve path handling when generating JSON files

cpiber:
Resolves #127 #282

Any build step for JavaScript files can now provide a mapping file, to allow the `po[t]` files to operate on the source files.

Mappings are simple JSON objects with source files as keys and one or an array of build files as values.

Currently this is done by a `--use-map` option, optionally with paths (delimited by `,`) specified. I'm open for better names for this option as well as better ways to specify multiple values (it doesn't seem like specifying the option multiple times works, like some programs support).

Example plugin (for webpack) to produce such maps: https://github.com/cpiber/webpack-bundle-output. I intend to integrate this plugin into `@wordpress/scripts` if this PR goes through, if that happens, that could be added to the README.


src/MakeJsonCommand.php
```diff
@@ -48,6 +48,9 @@ class MakeJsonCommand extends WP_CLI_Command {
 	 * [--pretty-print]
 	 * : Pretty-print resulting JSON files.
 	 *
+	 * [--use-map[=<paths>]]
+	 * : Whether to use a mapping file for the strings. Separate with `,` to specify multiple. Defaults to `build/map.json`.
```
swissspidy, line 52: I don't think we should default to any value here. This makes an assumption about the directory structure that won't fit for many projects.

Relatedly, would it perhaps make sense to allow JSON input such as `--input="{ \"src/index.js\": \"build/index.js\" }"` so one doesn't have to _have_ a map file?

src/MakeJsonCommand.php
```diff
@@ -63,6 +66,7 @@ class MakeJsonCommand extends WP_CLI_Command {
 	public function __invoke( $args, $assoc_args ) {
 		$purge           = Utils\get_flag_value( $assoc_args, 'purge', true );
 		$update_mo_files = Utils\get_flag_value( $assoc_args, 'update-mo-files', true );
+		$map_paths       = Utils\get_flag_value( $assoc_args, 'use-map', false );
```
swissspidy, line 69:
```suggestion
		$map_paths       = Utils\get_flag_value( $assoc_args, 'use-map', '' );
```

An empty string as the default value could make `build_map()` a bit simpler

src/MakeJsonCommand.php
```diff
@@ -127,14 +137,74 @@ public function __invoke( $args, $assoc_args ) {
...
+		}
+
+		$map = [];
+
+		if ( true === $paths ) {
+			$paths = [ 'build' . DIRECTORY_SEPARATOR . 'map.json' ];
+		} else {
+			$paths = array_unique( array_filter( explode( ',', $paths ) ) );
+		}
+		WP_CLI::debug( sprintf( 'Using files %s', implode( ', ', $paths ) ), 'make-json' );
```
swissspidy, line 158:
```suggestion
		WP_CLI::debug( sprintf( 'Using map files %s', implode( ', ', $paths ) ), 'make-json' );
```

src/MakeJsonCommand.php
```diff
@@ -127,14 +137,74 @@ public function __invoke( $args, $assoc_args ) {
...
+
+		foreach ( $paths as $path ) {
+			if ( ! file_exists( $path ) || is_dir( $path ) ) {
+				WP_CLI::warning( sprintf( 'File %s does not exist', $path ) );
```
swissspidy, line 162:
```suggestion
				WP_CLI::warning( sprintf( 'Map file %s does not exist', $path ) );
```

src/MakeJsonCommand.php
```diff
@@ -80,6 +84,12 @@ public function __invoke( $args, $assoc_args ) {
 			$destination = $args[1];
 		}
 
+		$map = $this->build_map( $map_paths );
+		if ( is_array( $map ) && empty( $map ) ) {
+			WP_CLI::warning( 'No valid keys found. No file was created.' );
+			return;
```
swissspidy, line 90: If an invalid map causes the script to bail early, this should probably use `WP_CLI::error()` to properly exit the script.

src/MakeJsonCommand.php
```diff
@@ -127,14 +137,74 @@ public function __invoke( $args, $assoc_args ) {
...
+
+		return $map;
+	}
+
 	/**
 	 * Splits a single PO file into multiple JSON files.
 	 *
 	 * @param string $source_file Path to the source file.
 	 * @param string $destination Path to the destination directory.
+	 * @param array  $map               Source to build file mapping.
```
swissspidy, line 204:
```suggestion
	 * @param array|null  $map               Source to build file mapping.
```

According to the code below and `build_map()`, it can also be `null`

Although we could simplify the code below if it were never `null`, but simply an empty `array` instead.

src/MakeJsonCommand.php
```diff
@@ -127,14 +137,74 @@ public function __invoke( $args, $assoc_args ) {
 		WP_CLI::success( sprintf( 'Created %d %s.', $result_count, Utils\pluralize( 'file', $result_count ) ) );
 	}
 
+	/**
+	 * Collect maps from paths, normalize and merge
+	 *
+	 * @param string|bool $paths argument. False to do nothing.
+	 * @return array|null Mapping array. Null if no maps specified.
```
swissspidy, line 144: As per my other comment, perhaps return an empty array if no maps specified?

src/MakeJsonCommand.php
```diff
@@ -147,6 +217,39 @@ protected function make_json( $source_file, $destination ) {
...
+					$references = array_merge( $references, $sources );
+				}
+				// and wrap to array
+				$references = array_map(
+					function ( $value ) {
+						return [ $value ];
+					},
+					$references
+				);
+			}
```
swissspidy, line 251: Perhaps extract this to a method?

swissspidy (commented):
Apologies for the long wait, this is really nice work! 🎉 

Just a left a few minor comments

src/MakeJsonCommand.php
cpiber: It would, but then I couldn't differentiate between argument not given and `--use-map=`

swissspidy: Is such differentiation important?

swissspidy:
By the way, nice work with https://github.com/cpiber/webpack-bundle-output too!

src/MakeJsonCommand.php
cpiber: Hmm, nice suggestion, but my problem is now how do I parse the argument? The first thing I thought of was just to use `json_decode`, but that means the user has to put quotes around the strings, are we fine with that?

src/MakeJsonCommand.php
cpiber: But if it were never null, we'd always have to have a map, since translations not in the map are dropped

src/MakeJsonCommand.php
cpiber: Actually, nevermind, what do you think of this regex:

```regex
(?:^|(?<=,))\s*([^{[}\]"',]*?)\s*(?:$|(?=,))
```

We could maybe use that to add quotes around the strings

swissspidy: `Utils\parse_shell_arrays()` already does most of the heavy lifting here. It calls `Utils\is_json()` and runs `json_decode()` if needed. So we don't need to do anything extra here.

Check out `MakePotCommand` and how the `headers` argument (which allows JSON) is parsed there.

src/MakeJsonCommand.php
cpiber: I think I was being unclear: We couldn't know whether the argument was given or not. Since we only want to transform the sources if the argument is given, it is.

src/MakeJsonCommand.php
cpiber: Thank you, I didn't know of that, I'll look into that!

src/MakeJsonCommand.php
```diff
@@ -56,13 +60,24 @@ class MakeJsonCommand extends WP_CLI_Command {
 	 *     # Create JSON files for my-plugin-de_DE.po and leave the PO file untouched.
 	 *     $ wp i18n make-json my-plugin-de_DE.po /tmp --no-purge
 	 *
+	 *     # Create JSON files with mapping
+	 *     $ wp i18n make-json languages --use-map=build/map.json
+	 *
+	 *     # Create JSON files with multiple mappings
+	 *     $ wp i18n make-json languages '--use-map=["build/map.json","build/map2.json"]'
```
swissspidy, line 67: Interesting, hadn't thought about that. To pass multiple files, I would have imagined `--use-map=build/map.json,build/map2.json` to be more intuitive. Is that also supported?

cpiber: No, because that would cause problems with passing objects.

Previously, when only paths were accepted, that was the way.

But not, it would not be possible to distinguish between the comma inside an object and the comma between files, at least not trivially. Hence my suggestion of a regex mentioned here https://github.com/wp-cli/i18n-command/pull/284#discussion_r714748833

swissspidy (approved):
Overall looks good.

Just one question came to mind right now

PR #302, opened 2022-01-17
Fix include logic where include path is subdirectory of excluded path.

shendy-a8c:
Fixes https://github.com/wp-cli/i18n-command/issues/172.

Changes in this PR:
- Fix `containsMatchingChildren()`, so $root_relative_path doesn't have trailing slash. [[#](https://github.com/shendy-a8c/i18n-command/blob/e52b2aa7ed2fdaa6186ef530e231e7847a54c97e/src/IterableCodeExtractor.php#L171)]. This is because [`$include` and `$exclude` are trailing-slash-trimmed](https://github.com/shendy-a8c/i18n-command/blob/e52b2aa7ed2fdaa6186ef530e231e7847a54c97e/src/IterableCodeExtractor.php#L219-L220), so if `$root_relative_path` isn't, [this clause](https://github.com/shendy-a8c/i18n-command/blob/e52b2aa7ed2fdaa6186ef530e231e7847a54c97e/src/IterableCodeExtractor.php#L178) won't be true as expected.
- When the file iterator gets to the included folder, which is a subfolder of an excluded folder, it will get [here](https://github.com/shendy-a8c/i18n-command/blob/e52b2aa7ed2fdaa6186ef530e231e7847a54c97e/src/IterableCodeExtractor.php#L242), so I created a new [if block](https://github.com/shendy-a8c/i18n-command/blob/e52b2aa7ed2fdaa6186ef530e231e7847a54c97e/src/IterableCodeExtractor.php#L236-L239) to handle that case.
- Unit test for this case: including a folder inside an excluded folder.

### DIFF
--- a/src/IterableCodeExtractor.php
+++ b/src/IterableCodeExtractor.php
@@ -168,6 +168,7 @@ trait IterableCodeExtractor {
 
 		/** @var string $root_relative_path */
 		$root_relative_path = str_replace( static::$dir, '', $dir->getPathname() );
+		$root_relative_path = self::trim_leading_slash( $root_relative_path );
 
 		foreach ( $matchers as $path_or_file ) {
 			// If the matcher contains no wildcards and the path matches the start of the matcher.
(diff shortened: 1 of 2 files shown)


src/IterableCodeExtractor.php
```diff
@@ -232,6 +233,11 @@ static function ( $file, $key, $iterator ) use ( $include, $exclude, $extensions
 						return static::containsMatchingChildren( $file, $include );
 					}
 
+					// Include directories that are excluded but include score is higher.
+					if ( 0 < $exclusion_score && $inclusion_score >= $exclusion_score && $iterator->hasChildren() ) {
```
swissspidy, line 237: Nit: no yoda conditions here.

```suggestion
					if ( $exclusion_score > 0 && $inclusion_score >= $exclusion_score && $iterator->hasChildren() ) {
```

shendy-a8c: Addressed in b079a92.

shendy-a8c:
@swissspidy I'm not rushing but just want to make sure if there is any action item on my side.

swissspidy (approved):

Issue #147, opened 2019-03-13
Support Blade Templates


connerbw:
Currently, this CLI command supports [PHP](https://github.com/wp-cli/i18n-command/blob/88f5b687f715979ba89bbc2e2075465a84ff88ac/src/MakePotCommand.php#L569) and [two](https://github.com/wp-cli/i18n-command/blob/88f5b687f715979ba89bbc2e2075465a84ff88ac/src/MakePotCommand.php#L572) [types](https://github.com/wp-cli/i18n-command/blob/88f5b687f715979ba89bbc2e2075465a84ff88ac/src/MakePotCommand.php#L582) of JavaScript extraction. Eventually, the code that does this leads back to / depends on: [Gettext\Extractors\PhpCode](https://github.com/oscarotero/Gettext/blob/master/src/Utils/PhpFunctionsScanner.php) and [ Gettext\Utils\JsFunctionsScanner](https://github.com/oscarotero/Gettext/blob/master/src/Utils/JsFunctionsScanner.php) 

`oscarotero/Gettext` supports [several other extractors](https://github.com/oscarotero/Gettext#extractors), one of them is Blade templates.

Our WordPress project uses [Blade templates](https://github.com/jenssegers/blade). The naming convention for Blade templates is: `/path/to/name.blade.php` The strings inside those files looks like: `{{ __( 'My English string', 'my-plugin-name' ) }}` The i18n-command finds, scans, and fails to extract anything meaningful from those files.

Please support Blade templates.  I'm willing to write a PR to do this if this issue isn't closed as a #wontfix.

Related: https://github.com/wp-cli/i18n-command/issues/52

swissspidy:
How are you currently extracting strings from your files? Do you perhaps have a link to your project?

The original goal was to make it easy to extract strings in a similar fashion to the old makepot script, which is only concernced with "regular" PHP files, and to add support for the new JS features.

connerbw:
Hi @swissspidy

> How are you currently extracting strings from your files? Do you perhaps have a link to your project?

The project is Pressbooks:
https://github.com/pressbooks/pressbooks/tree/dev/languages

We use wp-cli to generate the .pot source files. Note: [We run your branch for plurals.](https://github.com/wp-cli/i18n-command/pull/128) We do: 

    wp i18n make-pot . languages/<plugin>.pot --exclude="vendor,symbionts" --headers='{"Report-Msgid-Bugs-To":"https://github.com/pressbooks/<plugin>/"}'

[...]

PR #304, opened 2022-02-21
Add support for PHP-Blade files

strarsis:
This PR adds the code and (simple) tests for extracting translations strings from PHP-Blade template files.

Luckily the [`4.x` branch of the `gettext` project already contains a fully working PHP-Blade Extractor](https://github.com/php-gettext/Gettext/blob/4.x/src/Extractors/Blade.php), 
which can be used very similarly to the PHP extractor for `wp i18n`.

The `IterableCodeExtractor` file filter [had to be refactored](https://github.com/wp-cli/i18n-command/pull/304/files#diff-6dbbabe411870b35fac549f92d2c059e9cc16d2f0f85b37347393150a9f3e7c1) a bit to make it also support filtering for multiple file extensions (as it is the case with `blade.php` for PHP-Blade template files).

The ~~`illuminate/view`~~ `BladeOne` compiler for a standalone PHP-Blade compiler [had to be included as `composer` dependency](https://github.com/wp-cli/i18n-command/pull/304/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R15) as the `gettext` package doesn't require it on its own/expects a PHP-Blade compiler to be available. Edit: Because of current PHP version constraints by `i18n`, a release of another, standalone and well-maintained PHP-Blade compiler is used instead (`BladeOne`).

### Discussion
Both classes, `PhpCodeExtractor` and `BladeCodeExtractor`  duplicate some code, `$options` and `$functionsScannerClass`:
https://github.com/strarsis/i18n-command/blob/add-blade-extractor/src/PhpCodeExtractor.php#L13-L46
https://github.com/strarsis/i18n-command/blob/add-blade-extractor/src/BladeCodeExtractor.php#L13-L46

Side note: Putting these two aforementioned fields `$options` and `$functionsScannerClass` into a PHP `trait` doesn't work as PHP `trait`s can't override class fields (as I learned by trying to do so).
Both classes have to extend from a class from `gettext/gettext`, therefore a trivial way for letting both classes share these fields doesn't seem to exist, so I think those duplications (which are quite compact) can be tolerated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,8 +11,8 @@
         }
     ],
     "require": {
+        "eftec/bladeone": "3.52",
         "gettext/gettext": "^4.8",
-        "illuminate/view": "^8.83",
         "mck89/peast": "^1.13.11",
         "wp-cli/wp-cli": "^2.5"
     },
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
     ],
     "require": {
         "gettext/gettext": "^4.8",
+        "illuminate/view": "^8.83",
         "mck89/peast": "^1.13.11",
         "wp-cli/wp-cli": "^2.5"
     },
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -2911,6 +2911,52 @@ Feature: Generate a POT file of a WordPress project
       msgid "Bar"
       """
 
+  Scenario: Extract strings from a Blade-PHP file
+    Given an empty example-project directory
+    And a example-project/stuff.blade.php file:
+      """
+	  @php
+		__('Test');
+	  @endphp
+      @extends('layouts.app')
+	  
+	  @php(__('Another test.', 'domain'))
+
+      @section('content')
+        @include('partials.page-header')
+
+        @if (! have_posts())
+          <x-alert type="warning">
+            {!! __('Page not found.', 'foo-plugin') !!}
+          </x-alert>
+
+          {!! get_search_form(false) !!}
+        @endif
+      @endsection
+      """
+
+    When I try `wp i18n make-pot example-project result.pot --ignore-domain --debug`
+    Then STDOUT should be:
+      """
+      Success: POT file successfully generated!
+      """
+    And STDERR should contain:
+      """
+      No valid theme stylesheet or plugin file found, treating as a regular project.
+      """
+    And the result.pot file should contain:
+      """
+      msgid "Test"
+      """
+    And the result.pot file should contain:
+      """
+      msgid "Page not found."
+      """
+    And the result.pot file should contain:
+      """
+      msgid "Another test."
+      """
+
   Scenario: Custom package name
     Given an empty example-project directory
     And a example-project/stuff.php file:
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -2911,6 +2911,7 @@ Feature: Generate a POT file of a WordPress project
       msgid "Bar"
       """
 
+  @blade
   Scenario: Extract strings from a Blade-PHP file in a theme (ignoring domains)
     Given an empty foo-theme directory
     And a foo-theme/style.css file:
@@ -2967,6 +2968,7 @@ Feature: Generate a POT file of a WordPress project
       msgid "Another test."
       """
 
+  @blade
   Scenario: Extract strings from a Blade-PHP file in a theme
     Given an empty foo-theme directory
     And a foo-theme/style.css file:
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -2933,7 +2933,7 @@ Feature: Generate a POT file of a WordPress project
 		__('Test');
 	  @endphp
       @extends('layouts.app')
-	  
+
 	  @php(__('Another test.', 'some-other-domain'))
 
       @section('content')
@@ -2990,7 +2990,7 @@ Feature: Generate a POT file of a WordPress project
 		__('Test');
 	  @endphp
       @extends('layouts.app')
-	  
+
 	  @php(__('Another test.', 'some-other-domain'))
 
       @section('content')
@@ -3011,10 +3011,6 @@ Feature: Generate a POT file of a WordPress project
       """
       Theme stylesheet detected.
       Success: POT file successfully generated!
-      """
-    And the result.pot file should contain:
-      """
-      msgid "Test"
       """
     And the result.pot file should contain:
       """
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -2911,9 +2911,23 @@ Feature: Generate a POT file of a WordPress project
       msgid "Bar"
       """
 
-  Scenario: Extract strings from a Blade-PHP file
-    Given an empty example-project directory
-    And a example-project/stuff.blade.php file:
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -12,6 +12,7 @@
     ],
     "require": {
         "gettext/gettext": "^4.8",
+        "illuminate/view": "^8.83",
```
swissspidy, line 15: What's this for?

features/makepot.feature
```diff
@@ -2911,6 +2911,52 @@ Feature: Generate a POT file of a WordPress project
...
+          <x-alert type="warning">
+            {!! __('Page not found.', 'foo-plugin') !!}
+          </x-alert>
+
+          {!! get_search_form(false) !!}
+        @endif
+      @endsection
+      """
+
+    When I try `wp i18n make-pot example-project result.pot --ignore-domain --debug`
```
swissspidy, line 2938: An additional scenario for testing extraction without `--ignore-domain` would be helpful too.

features/makepot.feature
```diff
@@ -2911,6 +2911,52 @@ Feature: Generate a POT file of a WordPress project
...
+    Then STDOUT should be:
+      """
+      Success: POT file successfully generated!
+      """
+    And STDERR should contain:
+      """
+      No valid theme stylesheet or plugin file found, treating as a regular project.
```
swissspidy, line 2945: For the sake of the test a plugin or theme file could be added so this error doesn't appear

src/IterableCodeExtractor.php
```diff
@@ -262,6 +262,23 @@ static function ( $file, $key, $iterator ) use ( $include, $exclude, $extensions
...
+		$file_extension_separator = '.';
+
+		$filename = $file->getFilename();
+		$parts    = explode( $file_extension_separator, $filename, 2 );
+		if ( count( $parts ) <= 1 ) {
+			// if ever something goes wrong, fall back to SPL
+			return $file->getExtension();
+		}
+		return $parts[1];
+	}
```
swissspidy, line 280: PHPDoc blocks for these two methods and perhaps some more explicit test coverage in `IterableCodeExtractorTest` would be nice, so it's more obvious what they do.

src/MakePotCommand.php
```diff
@@ -609,6 +618,18 @@ protected function extract_strings() {
 				PhpCodeExtractor::fromDirectory( $this->source, $translations, $options );
 			}
 
+			if ( ! $this->skip_blade ) {
+				$options = [
+					// Extract 'Template Name' headers in theme files.
+					'wpExtractTemplates' => isset( $this->main_file_data['Theme Name'] ),
```
swissspidy, line 624: This is not relevant for Blade files; should be able to safely remove

swissspidy:
Thanks for opening this!

For context, Blade support was originally proposed in #147 but was more seen as out of scope. But this might be a reasonable addition after all.

strarsis:
@swissspidy: After adding a test that doesn't ignore the translation domains, I noticed an issue with the PHP function scanner. It probably needs some adjustments. 🐱

swissspidy:
> @swissspidy: After adding a test that doesn't ignore the translation domains, I noticed an issue with the PHP function scanner. It probably needs some adjustments. 🐱

What kind of issue?

> The `illuminate/view` for the standalone `Blade` compiler [had to be included as `composer` dependency](https://github.com/wp-cli/i18n-command/pull/304/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R15) as the `gettext` package doesn't require it on its own/expects a Blade compiler to be available.

`illuminate/view` v8.83.0 requires PHP 7.3, but this command still supports PHP 7.0, so that doesn't work. Unfortunately all older versions that would be compatible have security advisories.

strarsis:
@swissspidy: The `illuminate/view` - PHP `7.0` issue was solved by using the very well maintained `BladeOne` standalone Blade compiler/engine release `3.52` which is also supported by PHP `7.0`:
https://github.com/EFTEC/BladeOne/blob/3.52/composer.json#L26

swissspidy:
@schlessera WDYT about the PHP version requirement bump mentioned above (and this enhancement in general)?

tests/IterableCodeExtractorTest.php
```diff
@@ -203,4 +204,47 @@ public function test_can_include_file_in_symlinked_folder() {
...
+		return $file_get_extension_multi_method->invokeArgs( null, [ $file, $extensions ] );
+	}
+
+	public function test_gets_file_extensions_correctly() {
+		$this->assertEquals( static::file_get_extension_multi_invoke( new \SplFileObject( self::$base . 'foo/bar/foofoo/included.js' ) ), 'js' );
+		$this->assertEquals( static::file_get_extension_multi_invoke( new \SplFileObject( self::$base . 'foo-plugin/foo-plugin.php' ) ), 'php' );
+		$this->assertEquals( static::file_get_extension_multi_invoke( new \SplFileObject( self::$base . 'foo-theme/foo-theme-file.blade.php' ) ), 'blade.php' );
+	}
+
+	public function test_matches_file_extensions_correctly() {
```
swissspidy, line 232: Nit: could use a data provider here

schlessera:
@swissspidy, @strarsis The current PHP version requirement policy is the following:
"**Whenever WordPress Core raises the PHP minimum version requirement past WP-CLI, WP-CLI will stay at the old minimum requirement for at least 1 full year.**"
(see https://make.wordpress.org/cli/2019/01/15/wp-cli-php-requirements-strategy/)
This means  we still need to stay at PHP 5.6+ for now and are blocked to that minimum as long as WP Core itself does not bump it.

strarsis:
@schlessera: So even PHP 7.1 would be still too high for `i18n` then.

The good news is that the Blade Engine `BladeOne` release v3.52 (which doesn't have  used in this PR [supports PHP `5.6+`](https://github.com/EFTEC/BladeOne/blob/3.52/composer.json#L26)!

swissspidy:
This command already has a higher PHP version requirement AFAIK

schlessera:
Nope, it is still being tested against 5.6
Merge pull request #302 from shendy-a8c_fix_include-subfolder-of-excl… · wp  2022-03-11 at 10 23 22 AM
:

swissspidy (approved):

strarsis:
@swissspidy: Great that this was merged! I can install this WP CLI command from source, but it would be even better if it is published as a new release. Are there plans for a new release of `i18n` in the near future?

swissspidy:
We should be able to tag a new release soon, but usually that's done closer to the release of WP-CLI itself. In the meantime, `wp package install git@github.com:wp-cli/i18n-command.git` is the way to go.

strarsis:
@swissspidy: There is an issue with installing WP CLI i18n from the repository (`wp package install git@github.com:wp-cli/i18n-command.git`):
The new `wp i18n` uses the previous version of `IterableCodeExtractor` from the `wp` CLI `PHAR`, 
which prevents `.blade.php` files from being correctly filtered in/out.

Hopefully there is a clean release soon 🐱

PR #292, opened 2021-12-02
Add new field to translate in `theme.json`

oandregal:
Depends on https://github.com/WordPress/gutenberg/pull/37038

In https://github.com/WordPress/gutenberg/pull/37038 Gutenberg is adding a new field to translate for the font sizes stored within a `theme.json`. This PR adds it to the i18n command.

## Setup

- See https://make.wordpress.org/cli/handbook/contributions/pull-requests/#setting-up
- To run the behat tests you also need a working mysql locally:
  - Install the packages if you don't have them. In linux: `sudo apt install jq mysql-server mysql-client php-mysql`
  - `composer install && composer behat` => it didn't setup the database for me (mysql 8) so I did;
    - `sudo mysql -u root`
    - `CREATE DATABASE IF NOT EXISTS `wp_cli_test`;`
    - `CREATE USER 'wp_cli_test'@'localhost' IDENTIFIED WITH mysql_native_password BY 'password1';`
    - `GRANT ALL PRIVILEGES ON wp_cli_test.* TO "wp_cli_test"@"localhost"`

## How to test

- Check out this PR locally and cd to the directory.
- `composer install`
- Edit `ThemeJsonExtractor.php` by substituting the line 148 `$json = self::remote_get( self::THEME_JSON_SOURCE );` by `$json = '';`.
- Create a new directory called `foo-theme` and create a `theme.json` file within with the following contents:

```json
{
  "version": "2",
  "settings": {
    "typography": {
      "fontSizes": [
        { "slug": "small", "size": "12px", "name": "Small", "alias": "S" }
      ]
    }
  }
}
```

- Run `vendor/bin/wp i18n make-pot foo-theme`
- Verify that there's a `foo-theme/foo-theme.pot` file and that it has the name and alias strings:

```pot
#: ...

#: theme.json
msgctxt "Font size name"
msgid "Small"
msgstr ""

#: theme.json
msgctxt "Short font size name"
msgid "S"
msgstr ""
```

PR #306, opened 2022-03-10
Translate `title` field from `theme.json`

oandregal:
Related https://github.com/wp-cli/i18n-command/pull/254 https://github.com/wp-cli/i18n-command/pull/292

In https://github.com/WordPress/gutenberg/pull/39322 and https://github.com/WordPress/gutenberg/pull/39715 the Gutenberg plugin has added a new translatable field to `theme.json` files, which will be part of WordPress 6.0.

This PR does a few things:

- Adds the new field to the local i18n schema.
- Updates the code so top-level keys are picked up for translation.
- Updates the code so strings are translatable (in addition to leaf arrays, as it works now).

## How to test

Automated tests:

`composer behat -- features/makepot.feature`

Manual test:

- `composer install`
- Comment [this line](https://github.com/wp-cli/i18n-command/blob/main/src/ThemeJsonExtractor.php#L148) in the `ThemeJsonExtractor` class, so the command uses the local backup file (`assets/theme-i18n.json`).
- Create a new directory called `foo-theme` that contains a `theme.json` file with the following contents:

```json
{
    "title": "My style variation"
}
```

- Verify that there's a `foo-theme/foo-theme.pot` file and that the "My style variation" string is present as well as its context "Style variation name". You should see something like:

```
msgctxt "Style variation name"
msgid "My style variation"
msgstr ""
```

### DIFF
--- a/assets/theme-i18n.json
+++ b/assets/theme-i18n.json
@@ -1,4 +1,5 @@
 {
+	"name": "Name of the style variation",
 	"settings": {
 		"typography": {
 				"fontSizes": [
--- a/assets/theme-i18n.json
+++ b/assets/theme-i18n.json
@@ -1,5 +1,5 @@
 {
-	"title": "Style variation title",
+	"title": "Style variation name",
 	"settings": {
 		"typography": {
 				"fontSizes": [
--- a/assets/theme-i18n.json
+++ b/assets/theme-i18n.json
@@ -1,5 +1,5 @@
 {
-	"name": "Name of the style variation",
+	"title": "Style variation title",
 	"settings": {
 		"typography": {
 				"fontSizes": [
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -3591,7 +3591,7 @@ Feature: Generate a POT file of a WordPress project
       """
     And the foo-theme/foo-theme.pot file should contain:
       """
-      msgctxt "Style variation title"
+      msgctxt "Style variation name"
       msgid "My style variation"
       """
 
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -3501,7 +3501,6 @@ Feature: Generate a POT file of a WordPress project
       msgid "Notice"
       """
 
-
   Scenario: Skips theme.json file if skip-theme-json flag provided
     Given an empty foo-theme directory
     And a foo-theme/theme.json file:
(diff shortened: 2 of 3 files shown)


assets/theme-i18n.json
```diff
@@ -1,4 +1,5 @@
 {
+	"name": "Name of the style variation",
```
swissspidy, line 2:
```suggestion
	"name": "Style variation name",
```

This format is more in line with the other contexts

Of course you need to update https://github.com/WordPress/gutenberg/blob/f9c2838a0b8be94fcf4e51b0a831d4629eaf24d7/lib/compat/wordpress-5.9/theme-i18n.json as well.

oandregal:
@swissspidy this is ready for review.

swissspidy:
Change looks good, but note that Behat tests are now failing.

Is it on someone's radar to update `theme-i18n.json` in core as well?

oandregal:
The functional tests are failing because they're using the schema pulled from `https://develop.svn.wordpress.org/trunk/src/wp-includes/theme-i18n.json` to extract the strings, which doesn't have the `title` key yet. I'm preparing PRs for WordPress trunk and Gutenberg.

oandregal:
The changes to the i18n schema in Gutenberg and WordPress `trunk` have been deployed a couple of hours ago. However, the file at `https://develop.svn.wordpress.org/trunk/src/wp-includes/theme-i18n.json` doesn't yet have them despite it saying that it's at `Revision 53038` (the one that was [deployed](https://core.trac.wordpress.org/browser/trunk/src/wp-includes/theme-i18n.json?rev=53038)). Perhaps develop.svn.wordpress.org is something we only update nightly? I'll check tomorrow.

swissspidy:
https://develop.svn.wordpress.org/trunk/src/wp-includes/theme-i18n.json does have the `title` field now.

assets/theme-i18n.json
```diff
@@ -1,4 +1,5 @@
 {
+	"title": "Style variation title",
```
swissspidy, line 2:
```suggestion
	"title": "Style variation name",
```

@oandregal This is how it's in the schema. Can you confirm that `Style variation name` is the desired context? If so, you'd need to change it here and also in the Behat test.

oandregal: :+1:

PR #311, opened 2022-03-31
Look for `theme.json` files in the `styles` directory of a theme

oandregal:
This PR is a continuation of https://github.com/wp-cli/i18n-command/pull/306

It adds supports for looking for translatable strings in JSON files within the `styles` directory of a theme. Unlike the top-level `theme.json` file we don't lock any names in this case: any JSON file within the `styles` folder is a candidate we should consider.

## How to test

Automated tests:

`composer behat -- features/makepot.feature`

Manual test:

- `composer install`
- Create a new directory called `foo-theme` that contains a `styles` directory.
- Within the `styles` directory create a `my-style.json` file with the following contents:

```json
{
  "settings": {
    "color": {
      "palette": [ { "name": "My color", "slug": "slug", "color": "value" } ]
    }
  }
}
```

- Verify that there's a `foo-theme/foo-theme.pot` file and that the "My color" string is present as well as its context "Color name". You should see something like:

```
msgctxt "Color name"
msgid "My color"
msgstr ""
```

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -3621,3 +3621,35 @@ Feature: Generate a POT file of a WordPress project
       msgctxt "Color name"
       msgid "Black"
       """
+
+  Scenario: Extract strings from style variations
+    Given an empty foo-theme/styles directory
+    And a foo-theme/styles/my-style.json file:
+      """
+      {
+        "version": "1",
+        "settings": {
+          "blocks": {
+            "core/paragraph": {
+              "color": {
+                "palette": [
+                  { "slug": "black", "color": "#000000", "name": "Black" }
+                ]
+              }
+            }
+          }
+        }
+      }
+      """
+    
+    When I try `wp i18n make-pot foo-theme`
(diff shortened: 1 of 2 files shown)


oandregal:
cc @swissspidy this is the continuation of https://github.com/wp-cli/i18n-command/pull/306

oandregal:
The behat tests are now throwing the following warning (hence they fail):

> PHP Warning:  Invalid argument supplied for foreach() in src/ThemeJsonExtractor.php on line 209

This warning also happens in `main`.

The reason for this warning is that the i18n schema used to translate `theme.json` files ([see](https://develop.svn.wordpress.org/trunk/src/wp-includes/theme-i18n.json)) has been updated this morning. However, the code that interprets it has not been updated it yet: https://github.com/wp-cli/i18n-command/pull/306 needs to be merged to fix that warning.

swissspidy:
Merged #306 now

swissspidy (approved):

PR #312, opened 2022-04-05
Extract pattern metadata (title & description)

oandregal:
In https://github.com/WordPress/gutenberg/pull/36751 Gutenberg added the ability to automatically register patterns that live in the `/patterns` directory. These patterns provide metadata via headers in the PHP file as in:

```php
<?php
/**
 * Title: My heading
 * Description: My pattern description.
 * Slug: my-theme/my-heading
 * Categories: text
 */
```

From the existing metadata, we need the `Title` to be translatable. This PR adds support for looking up those patterns and extracting the Title string into the pot file.

## How to test

- Pull this PR and create a `foo-theme` directory at the top-level, which contains the following files:

`style.css` with at least these contents:

```css
/*
Theme Name: foo theme
*/
```

`patterns/my-pattern.php` with at least these contents:

```php
<?php

/**
 * Title: My pattern title.
 * Description: My pattern description.
 */
```

- Execute

```sh
composer install
vendor/bin/wp i18n make-pot foo-theme
```

- Verify that a new `foo-theme/foo-theme.pot` file has been generated and that contains the pattern title and the theme name. You should get something along these lines:

```po
# some other file header metadata

#. Theme Name of the theme
msgid "foo theme"
msgstr ""

#: patterns/my-pattern.php
msgctxt "Pattern title"
msgid "My pattern title."
msgstr ""

#: patterns/my-pattern.php
msgctxt "Pattern description"
msgid "My pattern description."
msgstr ""
```

### DIFF
--- a/src/IterableCodeExtractor.php
+++ b/src/IterableCodeExtractor.php
@@ -67,7 +67,13 @@ trait IterableCodeExtractor {
 			}
 
 			if ( ! empty( $options['wpExtractPatterns'] ) ) {
-				$headers = MakePotCommand::get_file_data_from_string( $string, [ 'Title' => 'Title', 'Description' => 'Description' ] );
+				$headers = MakePotCommand::get_file_data_from_string(
+					$string,
+					[
+						'Title'       => 'Title',
+						'Description' => 'Description',
+					]
+				);
 
 				if ( ! empty( $headers['Title'] ) ) {
 					$translation = new Translation( '', $headers['Title'] );
--- a/src/IterableCodeExtractor.php
+++ b/src/IterableCodeExtractor.php
@@ -24,7 +24,7 @@ trait IterableCodeExtractor {
 	 *     Optional. An array of options passed down to static::fromString()
 	 *
 	 *     @type bool  $wpExtractTemplates Extract 'Template Name' headers in theme files. Default 'false'.
-	 *     @type bool  $wpExtractPatterns  Extract 'Title' headers in pattern files. Default 'false'.
+	 *     @type bool  $wpExtractPatterns  Extract 'Title' and 'Description' headers in pattern files. Default 'false'.
 	 *     @type array $restrictFileNames  Skip all files which are not included in this array.
 	 * }
 	 * @return null
--- a/src/IterableCodeExtractor.php
+++ b/src/IterableCodeExtractor.php
@@ -76,14 +76,14 @@ trait IterableCodeExtractor {
 				);
 
 				if ( ! empty( $headers['Title'] ) ) {
-					$translation = new Translation( 'Title of the pattern', $headers['Title'] );
+					$translation = new Translation( 'Pattern title', $headers['Title'] );
 					$translation->addReference( $options[ 'file' ] );
 
 					$translations[] = $translation;
 				}
 
 				if ( ! empty( $headers['Description'] ) ) {
-					$translation = new Translation( 'Description of the pattern', $headers['Description'] );
+					$translation = new Translation( 'Pattern description', $headers['Description'] );
 					$translation->addReference( $options[ 'file' ] );
 
 					$translations[] = $translation;
--- a/src/IterableCodeExtractor.php
+++ b/src/IterableCodeExtractor.php
@@ -76,15 +76,15 @@ trait IterableCodeExtractor {
 				);
 
 				if ( ! empty( $headers['Title'] ) ) {
-					$translation = new Translation( '', $headers['Title'] );
-					$translation->addExtractedComment( 'Title of the pattern' );
+					$translation = new Translation( 'Title of the pattern', $headers['Title'] );
+					$translation->addReference( $options[ 'file' ] );
 
 					$translations[] = $translation;
 				}
 
 				if ( ! empty( $headers['Description'] ) ) {
-					$translation = new Translation( '', $headers['Description'] );
-					$translation->addExtractedComment( 'Description of the pattern' );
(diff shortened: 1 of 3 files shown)


src/MakePotCommand.php
```diff
@@ -618,6 +618,17 @@ protected function extract_strings() {
 				PhpCodeExtractor::fromDirectory( $this->source, $translations, $options );
 			}
 
+			if ( ! $this->skip_php ) {
+				$options = [
+					// Extract 'Title' headers from pattern files.
+					'wpExtractPatterns' => isset( $this->main_file_data['Theme Name'] ),
```
oandregal, line 624: `isset( $this->main_file_data['Theme Name'] )` this means we only look for patterns in a "theme" directory: a directory that contains theme metadata in its `style.css`.

src/IterableCodeExtractor.php
```diff
@@ -65,6 +66,17 @@ public static function fromFile( $file_or_files, Translations $translations, arr
 				}
 			}
 
+			if ( ! empty( $options['wpExtractPatterns'] ) ) {
+				$headers = MakePotCommand::get_file_data_from_string( $string, [ 'Title' => 'Title' ] );
```
oandregal, line 70: Need to check what other metadata we need to translate. According to https://github.com/WordPress/gutenberg/pull/36751/files#diff-ef23af8c1b0d6931bd6bbedd91dcfa92b5537a438c1c0c45499b8efdf2c91546R84 we may need to consider `Description` as well.

swissspidy: _Description_ definitely sounds like something user-facing that needs i18n :-)

swissspidy:
> From the existing metadata, we need the `Title` to be translatable

For this to be actually useful, Gutenberg should also actually translate the title. I don't see any `$pattern_data['title'] = translate_with_gettext_context( $pattern_data['title'], 'Title of the pattern' );` call in https://github.com/WordPress/gutenberg/pull/36751

oandregal:
> For this to be actually useful, Gutenberg should also actually translate the title. I don't see any $pattern_data['title'] = translate_with_gettext_context( $pattern_data['title'], 'Title of the pattern' ); 

Yeah, I need to follow-up with that as well :)

src/IterableCodeExtractor.php
oandregal: Description added at https://github.com/wp-cli/i18n-command/pull/312/commits/cc4b02b4e3de8dde18e9f5f91d3c1e716c99fab7

oandregal:
I've prepared the Gutenberg PR that uses this data at https://github.com/WordPress/gutenberg/pull/40047 but I'm having some trouble to get the translated strings.

src/IterableCodeExtractor.php
```diff
@@ -65,6 +66,30 @@ public static function fromFile( $file_or_files, Translations $translations, arr
...
+				if ( ! empty( $headers['Title'] ) ) {
+					$translation = new Translation( 'Title of the pattern', $headers['Title'] );
+					$translation->addReference( $options[ 'file' ] );
+
+					$translations[] = $translation;
+				}
+
+				if ( ! empty( $headers['Description'] ) ) {
+					$translation = new Translation( 'Description of the pattern', $headers['Description'] );
+					$translation->addReference( $options[ 'file' ] );
```
swissspidy, line 87: Has there been consensus on what to use as the translation context for these new strings?

For `block.json` we use `block title` and `block description` for example, and for `theme.json` `Style variation name`.

So `Pattern title` and `Pattern description` would be more in line with that.

Thoughts?

oandregal: Updated at https://github.com/wp-cli/i18n-command/pull/312/commits/ffa87027ac9ff724d74e454d3d193bfcd7b8903e

swissspidy:
Pending some tests this looks good to go! 👍

oandregal:
> Pending some tests this looks good to go! +1

Tests would need to wait until tomorrow morning :)

oandregal:
@swissspidy pushed the test.

src/MakePotCommand.php
```diff
@@ -618,6 +618,17 @@ protected function extract_strings() {
 				PhpCodeExtractor::fromDirectory( $this->source, $translations, $options );
 			}
 
+			if ( ! $this->skip_php ) {
+				$options = [
+					// Extract 'Title' headers from pattern files.
```
swissspidy, line 623:
```suggestion
					// Extract 'Title' and 'Description' headers from pattern files.
```

src/MakePotCommand.php
```diff
@@ -618,6 +618,17 @@ protected function extract_strings() {
 				PhpCodeExtractor::fromDirectory( $this->source, $translations, $options );
 			}
 
+			if ( ! $this->skip_php ) {
+				$options = [
+					// Extract 'Title' and 'Description' headers from pattern files.
+					'wpExtractPatterns' => isset( $this->main_file_data['Theme Name'] ),
+					'include'           => array_merge( $this->include, array( 'patterns' ) ),
```
ocean90, line 625: @swissspidy Do you know why this hardcodes `patterns` here? Since these are PHP files they should be included by default. Now you explicitly have to exclude the `patterns` when you for example are using `--include=wp-admin/*`. to only include wp-admin strings.

swissspidy: I suppose the idea was to make this more efficient and only scan these files when needed.

But you're right that this is already covered by the call above. I'll create a PHP to combine these.

swissspidy: #320

PR #313, opened 2022-04-30
Prefix JSON files with text domain if needed

### LINKED ISSUE #197: make-json: generated file missing theme textdomain prefix
## Bug Report

I'll keep it brief – might hopefully save someones time. ;)

**Translations folder** in a theme – just as expected by `load_theme_textdomain()`:
```
languages/
    theme-slug.pot
    de_DE.po
    de_DE.mo
```

**CLI Command**
```
wp i18n make-json languages/*.po --no-purge
```

**The result**
```
languages/
    theme-slug.pot
    de_DE.po
    de_DE.mo
    languages/de_DE-a0623ef93657e56554c5ddda522ed690.json
```

The translation will not load.  
`wp_set_script_translations()` expects the file to live in `{textdomain}-{locale}-{md5}.json`.  
After renaming it accordingly the translation loads.

swissspidy:
Fixes #197

### DIFF
--- a/src/MakeJsonCommand.php
+++ b/src/MakeJsonCommand.php
@@ -242,7 +242,7 @@ class MakeJsonCommand extends WP_CLI_Command {
 		$domain = $translations->getDomain();
 
 		if ( $domain && 0 !== strpos( $base_file_name, $domain ) ) {
-			$base_file_name = "$domain-$base_file_name";
+			$base_file_name = "{$domain}-{$base_file_name}";
 		}
 
 		foreach ( $translations as $translation ) {
--- a/src/MakeJsonCommand.php
+++ b/src/MakeJsonCommand.php
@@ -197,7 +197,7 @@ class MakeJsonCommand extends WP_CLI_Command {
 			$key_num             = count( $json );
 			// normalize contents to string[]
 			$json = array_map(
-				static function ($value ) {
+				static function ( $value ) {
 					if ( is_array( $value ) ) {
 						$value = array_values( array_filter( $value, 'is_string' ) );
 						if ( ! empty( $value ) ) {
--- a/src/MakeJsonCommand.php
+++ b/src/MakeJsonCommand.php
@@ -132,12 +132,11 @@ class MakeJsonCommand extends WP_CLI_Command {
 
 					if ( $update_mo_files ) {
 						$file_basename    = basename( $file->getFilename(), '.po' );
-						$destination_file = "{$destination}/{$file_basename}.mo";
+						$destination_file = "$destination/$file_basename.mo";
 
 						$translations = Translations::fromPoFile( $file->getPathname() );
 						if ( ! $translations->toMoFile( $destination_file ) ) {
-							WP_CLI::warning( "Could not create file {$destination_file}" );
-							continue;
+							WP_CLI::warning( "Could not create file $destination_file" );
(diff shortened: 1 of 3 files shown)


src/MakeJsonCommand.php
```diff
@@ -132,12 +132,11 @@ public function __invoke( $args, $assoc_args ) {
 
 					if ( $update_mo_files ) {
 						$file_basename    = basename( $file->getFilename(), '.po' );
-						$destination_file = "{$destination}/{$file_basename}.mo";
+						$destination_file = "$destination/$file_basename.mo";
```
schlessera, line 135: Why remove the curly braces syntax here? That has been the syntax we're using across all repos so far...

src/MakeJsonCommand.php
```diff
@@ -240,12 +239,18 @@ protected function make_json( $source_file, $destination, $map ) {
 
 		$base_file_name = basename( $source_file, '.po' );
 
-		foreach ( $translations as $index => $translation ) {
+		$domain = $translations->getDomain();
+
+		$base_file_name_with_domain = $domain && 0 !== strpos( $base_file_name, $domain ) ?
```
schlessera, line 244: This needs parentheses to denote the intention clearly.
```suggestion
		$base_file_name_with_domain = ( $domain && 0 !== strpos( $base_file_name, $domain ) ) ?
```

schlessera (changes requested):

src/MakeJsonCommand.php
swissspidy: It was an IDE recommendation to remove unneeded parentheses.

Happy to add them back, since I shouldn't have included unrelated changes here anyway.

Is there perhaps a lint rule that could enforce this kinda style?

src/MakeJsonCommand.php
swissspidy: It was an IDE recommendation to remove unneeded curly braces.

Happy to add them back, since I shouldn't have included unrelated changes here anyway.

Is there perhaps a lint rule that could enforce this kinda style?

src/MakeJsonCommand.php
schlessera: The parentheses are technically not needed because of the precedence rules of the different operators.

However, given this specific construct, I had a hard time reading it correctly in its raw form. I had initially thought there was a bug with an inadvertent assignment.

swissspidy: Rewrote it now without a ternary

src/MakeJsonCommand.php
swissspidy: Reverted the changes now

schlessera (changes requested):

schlessera (approved):

drzraf:
I'm retrieving my PO files from Weblate using `wlc`: these are missing a proper `Domain` header and still, I'd expect `make-json` to be able to deal with that (eg with a `--domain` flag).
Not sure what workflow expectations are adequate, but for me, it makes sense to give control over a specific `domain` setting at the `make-json` step.

swissspidy:
@drzraf Please open a new issue for your enhancement request instead of commenting on an old merged PR. Thanks!

I also suggest reaching out to your PO file provider to include the Domain header, as it‘s generally a good practice.

PR #315, opened 2022-05-02
Add missing docs for make-mo command

### LINKED ISSUE #314: No example, huh!?
You provide an example for the make-pot command but not for the make-mo command. Why is that?

Now, I know that I COULD probably spend an hour or two trying to 'guess' what would work ... but why on earth be so inconsistent when creating documentation?

swissspidy:
Fixes #314

### DIFF
--- a/.github/settings.yml
+++ b/.github/settings.yml
@@ -26,3 +26,5 @@ labels:
     color: c5def5
   - name: command:i18n-make-json
     color: c5def5
+  - name: command:i18n-make-mo
+    color: c5def5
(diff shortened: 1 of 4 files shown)


wojsmol:
Direct changes in `README.md` are not necessary. This will be handled by [regenerate readme workflow](https://github.com/wp-cli/i18n-command/blob/main/.github/workflows/regenerate-readme.yml) after PR meagre.

swissspidy:
@wojsmol I know, but that just creates another PR adding noise. This saves the time :)

schlessera (approved):

PR #316, opened 2022-05-02
Add update-po command

### LINKED ISSUE #278: Feature Request: equivalent command for msgmerge
## Feature Request

As far as I know, `wp-cli i18n` doesn't provide an equivalent for gettext's `msgmerge` command. This command is really useful for updating existing `po` files by merging a `pot` template created with `make-pot`. So, for example, if I add a new `__` call in my plugin like this:
```php
$newtranslation = __('Translate me please', 'domain');
```
I can update the `messages.pot` file with `make-pot`, but I have to update existing `po` files with the new translation by hand.

To get around this, I've been calling `msgmerge` directly to update my `po` files:
```sh
find ./languages -name "*.po" | xargs -I % msgmerge -o % % languages/messages.pot
```
This merges the new translations of `messages.pot` into all existing `po` files in the `languages` directory. It would be great if the `i18n` command had a similar `update-pos` or `merge-pot` command to accomplish the same thing.

swissspidy:
Adds a new `wp i18n update-po` command that works similarly to [msgmerge](https://www.gnu.org/software/gettext/manual/html_node/msgmerge-Invocation.html)

Fixes #278

### DIFF
--- a/.github/settings.yml
+++ b/.github/settings.yml
@@ -26,3 +26,5 @@ labels:
     color: c5def5
   - name: command:i18n-make-json
     color: c5def5
+  - name: command:i18n-update-po
+    color: c5def5
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,8 @@
         "commands": [
             "i18n",
             "i18n make-pot",
-            "i18n make-json"
+            "i18n make-json",
+            "i18n update-po"
         ]
     },
     "autoload": {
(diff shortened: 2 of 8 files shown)


schlessera (approved):

strarsis:
@swissspidy; @schlessera: With the new PHP-Blade support, `theme.json` related fixes and now the `update-po` command, 
wouldn't it be a great time for a new release of `wp i18n`? Currently I use the `wp i18n` directly from GitHub, but installing and using it as-is from official release would make things much easier.

Issue #119, opened 2018-12-09
File name BEFORE the Warning & full path


kkmuffme:
It would be great if the i18n could include the full path to the files in Warnings & also put the file first, as this is more common among tools (eslint, php-cs-fixer,...) so easier to integrate into existing build pipelines.

Instead of this:
```
Warning: The string ""%s" added to cart" contains placeholders but has no "translators:" comment to clarify their meaning. (example.php:593)
```
we should see this:
```
/var/wp-content/plugins/example/example.php:593: Warning: The string ""%s" added to cart" contains placeholders but has no "translators:" comment to clarify their meaning.
```

Advantage:
- no need to change lots of build pipelines, bc the format is (almost) universal
- full path is better than only file name (especially when automating things)

swissspidy:
Great suggestion, thanks! Totally makes sense.

I wanted to improve the audit part for some time now anyway, see #39.

swissspidy:
Related: #167

elenachavdarova:
Pull request: https://github.com/wp-cli/i18n-command/pull/361

PR #318, opened 2022-05-02
Improve warnings for strings with different comments

### LINKED ISSUE #227: Translators comment audit - tell me in which files the different comment is
I have 1000s of files (from different projects that share a text-domain since its the same theme and texts just with slightly different design) with this string:
`'Hi %s'`

I get the following error:
Warning: The string "Hi %s," has 3 different translator comments. (some/file.php)

**Since wp-cli knows in which these different translator comments are, could you please also output this info to the user?**
Since I have about 7 of these warnings, it would take weeks to sift through all files and check where/what is different, since sometimes strings are on the same line, sometimes on the previous,...

swissspidy:
Lists all different comments attached to the string after the warning.

Example output:

```
$ wp i18n make-pot foo-plugin
Plugin file detected.
Warning: The string "Hello World" has 2 different translator comments. (foo-plugin.php:7)
translators: Translators 1!
Translators: Translators 2!
Success: POT file successfully generated!
```

Fixes #227

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -714,7 +714,7 @@ Feature: Generate a POT file of a WordPress project
       """
     And STDERR should contain:
       """
-      Warning: The string "Hello World" has 2 different translator comments. (foo-plugin.php:7)
+      Warning: The string "Hello World" has 2 different translator comments. (foo-plugin.php:7, foo-plugin.php:10)
       """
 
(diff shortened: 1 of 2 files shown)


swissspidy:
Related: #119

schlessera (approved):

PR #319, opened 2022-05-02
Update and combine `theme.json` and `block.json` extractors

### LINKED ISSUE #287: Use i18n schema for block.json from WordPress Core
## Feature Request

 

`BlockExtractor` is using a hardcoded list of fields to translate from `block.json` file:

https://github.com/wp-cli/i18n-command/blob/ea8a6b3ab19c490798b420beabf06ece40b8e708/src/BlockExtractor.php#L46-L66

Replicate the approach used for `ThemeJsonExtractor` that uses schema for `theme.json` located in WordPress core codebsase:

https://github.com/wp-cli/i18n-command/blob/ea8a6b3ab19c490798b420beabf06ece40b8e708/src/ThemeJsonExtractor.php#L106-L125

The schema for `block.json` is located at https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/block-i18n.json.

swissspidy:
Uses the `block-i18n.json` schema file to determine which fields are translatable.

**Enhancement:** Adds support for translating block variation fields. Apparently these were not extracted before...

Fixes #287


swissspidy:
cc @gziolo

swissspidy:
Looks like the existing theme i18n schema parser does not support  cases like 

```json
  "keywords": [ "block keyword" ],
```

as seen in `block.json`

So that needs some fiddling to enable support and make the tests pass

swissspidy:
@schlessera I'm running into some fatal errors here on PHP < 7.4. Perhaps you have some ideas here? Not sure which one of my changes could have caused that...

schlessera:
@swissspidy The problem was the use of `private` access in the `IterableCodeExtractor` trait. That worked fine until you extended a class that was using that trait.

swissspidy:
Ah 💡 🤦  Thanks!

gziolo:
I think it's a good idea to use the file from `wordpress-develop` for now. I'm still not sure whether using the canonical URL similar to https://schemas.wp.org/trunk/block.json as discussed in the parent issue would work so nicely as it can be ahead of WordPress core.

Anyway, it looks like this PR has everything covered. Nice work, thank you for bringing i18n schemas up to date ❤️

src/JsonSchemaExtractor.php
```diff
@@ -0,0 +1,180 @@
...
+		}
+
+		$headers  = [ 'Content-type: application/json' ];
+		$options  = [ 'halt_on_error' => false ];
+		$response = Utils\http_request( 'GET', $url, null, $headers, $options );
+
+		if (
+			! $response->success
+			|| 200 > (int) $response->status_code
+			|| 300 <= $response->status_code
```
schlessera, line 171:
```suggestion
			|| 300 <= (int) $response->status_code
```

src/JsonSchemaExtractor.php
```diff
@@ -0,0 +1,180 @@
...
+		) {
+			WP_CLI::debug( "Failed to download from URL $url", 'make-pot' );
```
schlessera, line 173:
```suggestion
			WP_CLI::debug( "Failed to download from URL {$url}", 'make-pot' );
```

src/JsonSchemaExtractor.php
```diff
@@ -0,0 +1,180 @@
...
+
+		return $file_structure;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public static function fromString( $string, Translations $translations, array $options = [] ) {
+		$file = $options['file'];
+		WP_CLI::debug( "Parsing file $file", 'make-pot' );
```
schlessera, line 81:
```suggestion
		WP_CLI::debug( "Parsing file {$file}", 'make-pot' );
```

src/JsonSchemaExtractor.php
```diff
@@ -0,0 +1,180 @@
...
+	/**
+	 * Fallback block-18n.json file path.
+	 *
+	 * @var string
+	 */
+	const BLOCK_JSON_FALLBACK = __DIR__ . '/../assets/block-i18n.json';
+
+	protected static $schema_cache = [];
+
+	/**
```
schlessera, line 43:
```suggestion
	/**
	 * Load the i18n from a remote URL or fall back to a local schema in case of an error.
```

src/JsonSchemaExtractor.php
```diff
@@ -0,0 +1,180 @@
...
+	const BLOCK_JSON_SOURCE = 'https://develop.svn.wordpress.org/trunk/src/wp-includes/block-i18n.json';
+
+	/**
+	 * Fallback block-18n.json file path.
+	 *
+	 * @var string
+	 */
+	const BLOCK_JSON_FALLBACK = __DIR__ . '/../assets/block-i18n.json';
+
+	protected static $schema_cache = [];
```
schlessera, line 41:
```suggestion
	/**
	 * Static cache for the remote schema files.
	 *
	 * @var array<string, string>
	 */
	protected static $schema_cache = [];
```

schlessera (changes requested):

schlessera (approved):

PR #350, opened 2023-02-02
Fix unclosed quote

cliffordp:
Not sure if it requires being surrounded by quotes so PR might need an adjustment in the other direction

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ if the source directory is detected as either a plugin or theme.
 
 	[--exclude=<paths>]
 		Comma-separated list of files and paths that should be skipped for string extraction.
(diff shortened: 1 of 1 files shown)


danielbachhuber (approved):
This works fine without the quote. Thanks @cliffordp !

swissspidy:
@danielbachhuber @cliffordp the readme is generated from the source code, so it needs to be fixed there instead. otherwise the chsnge gets overriddem with the next sync

danielbachhuber:
@swissspidy Oops, I thought this was a change to the command. I've updated the command with https://github.com/wp-cli/i18n-command/commit/9d09e903e96e9047336da9cb3451aa46d1586070

Thanks for the catch!

PR #351, opened 2023-02-05
Add more standard 'excludes' values

cliffordp:


src/MakePotCommand.php
```diff
@@ -47,7 +47,7 @@ class MakePotCommand extends WP_CLI_Command {
 	/**
 	 * @var array
 	 */
-	protected $exclude = [ 'node_modules', '.git', '.svn', '.CVS', '.hg', 'vendor', 'Gruntfile.js', 'webpack.config.js', '*.min.js' ];
+	protected $exclude = [ 'node_modules', '.*', '_*', 'vendor', 'Gruntfile.js', 'webpack.config.js', '*.min.js', 'test', 'tests' ];
```
danielbachhuber, line 50: `'.*', '_*'` seem a little broad. Could you share more detail about which folders you're solving for?

cliffordp: @danielbachhuber `.cache,.github,.parcel-cache,_entry,_test,_temp`

danielbachhuber: @cliffordp Makes sense. I think it'd be better to list those explicitly for now, if it's alright with you.

cliffordp: Are there any examples of dot files that should be processed, even in edge cases?
Is there an 'include' arg in the MakePot command to override any built-in excludes?

danielbachhuber: > Are there any examples of dot files that should be processed, even in edge cases?
> Is there an 'include' arg in the MakePot command to override any built-in excludes?

I don't have any specific examples off the top of my head. Given this is WordPress, I'm sure there's a way for this to be a backwards compatibility break for someone.

swissspidy: Yes there is an `--include` arg.

It's probably fine to exclude hidden directories by default, but not sure about the `_` prefix. That seems a bit broad indeed.

cliffordp: better now or you don't want `_entry` either?

danielbachhuber: @cliffordp Could you include some functional tests too? [Here's the file they should go in](https://github.com/wp-cli/i18n-command/blob/main/features/makepot.feature) and here is some [guidance on best practices](https://make.wordpress.org/cli/handbook/contributions/pull-requests/#functional-tests), if it's helpful.

Here's some prior art you can steal from: https://github.com/wp-cli/dist-archive-command/pull/61/files

swissspidy: But why `_entry`? Is that a common folder that many projects have and that should most likely never checked? It just seems very project-specific

cliffordp: @danielbachhuber not sure I have time to learn how to do that. I searched for `.git` since that was an existing excluded item, but I didn't see it.

@swissspidy removed

danielbachhuber: @cliffordp Sounds good. I added some tests with a58468b

@swissspidy I'm happy with this if you are.

swissspidy (approved):

PR #361, opened 2023-04-07
119 Filename before the warning and full path to it

elenachavdarova:
Included full path to the file and moved the path in the start of the warning.

Fixes: https://github.com/wp-cli/i18n-command/issues/119

### DIFF
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -282,7 +282,7 @@ class MakePotCommand extends WP_CLI_Command {
 	public function __invoke( $args, $assoc_args ) {
 		$this->handle_arguments( $args, $assoc_args );
 
-		$translations = $this->extract_strings();
+		$translations = $this->extract_strings( $args );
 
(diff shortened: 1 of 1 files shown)


danielbachhuber (changes requested):
Thanks for the pull request, @elenachavdarova !

Can you update the existing functional tests so they pass?

Also, what do you think about using a relative path instead of the full path? I think a relative path (relative to the directory they're in) might be a little more developer-friendly.

elenachavdarova:
Hello @danielbachhuber,
Thank you for your reply. For the full path - my opinion is that relative is better as well, however on the issue this PR is for - https://github.com/wp-cli/i18n-command/issues/119 the request is for full path (especially when automating). This is why I have set this up this way.

As for the funcitonal tests - I checked the errors upon the failed tests and I am not sure how to include the full path (as it is a tmp one) to the expected output. I have updated the PR with file name moved to the start of the error.

PR #363, opened 2023-05-05
Add `wp i18n make-php` command

swissspidy:
The WordPress core performance team is working on the Performant Translations feature project, which aims to speed up translations by using PHP files instead of MO files. The goal is to get it ready for WordPress 6.5.

While the idea is to have translate.wordpress.org ship these PHP files in language packs, some developers might be using WP-CLI for their i18n workflow, especially locally. For them it would be beneficial to be able to quickly generate the necessary translation files.

**Additional Context**

* https://make.wordpress.org/core/2023/09/05/call-for-testing-performant-translations/
* https://wordpress.org/plugins/performant-translations/
* https://github.com/swissspidy/performant-translations
* https://github.com/WordPress/wordpress-develop/pull/5306
* https://github.com/GlotPress/GlotPress/pull/1626

### DIFF
--- a/features/makephp.feature
+++ b/features/makephp.feature
@@ -126,7 +126,7 @@ Feature: Generate PHP files from PO files
       """
     And the foo-plugin/foo-plugin-de_DE.php file should contain:
       """
-      'messages'=>[''=>['Foo Plugin'=>[0=>'Foo Plugin']]]
+      'messages'=>[''=>['Foo Plugin'=>['Foo Plugin']]]
       """
 
   Scenario: Does include translations
@@ -163,5 +163,5 @@ Feature: Generate PHP files from PO files
     And the return code should be 0
     And the foo-plugin/foo-plugin-de_DE.php file should contain:
       """
-      'messages'=>[''=>['Foo Plugin'=>[0=>'Bar Plugin']]]
+      'messages'=>[''=>['Foo Plugin'=>['Bar Plugin']]]
       """
--- a/features/makephp.feature
+++ b/features/makephp.feature
@@ -1,0 +1,158 @@
+Feature: Generate PHP files from PO files
+
+  Background:
+    Given an empty directory
+
+  Scenario: Bail for invalid source directories
+    When I try `wp i18n make-php foo`
+    Then STDERR should contain:
+      """
+      Error: Source file or directory does not exist!
+      """
+    And the return code should be 1
+
+  Scenario: Uses source folder as destination by default
+    Given an empty foo-plugin directory
+    And a foo-plugin/foo-plugin-de_DE.po file:
+      """
+      # Copyright (C) 2018 Foo Plugin
+      # This file is distributed under the same license as the Foo Plugin package.
+      msgid ""
+      msgstr ""
+      "Project-Id-Version: Foo Plugin\n"
+      "Report-Msgid-Bugs-To: https://wordpress.org/support/plugin/foo-plugin\n"
+      "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+      "Language-Team: LANGUAGE <LL@li.org>\n"
+      "Language: de_DE\n"
+      "MIME-Version: 1.0\n"
+      "Content-Type: text/plain; charset=UTF-8\n"
+      "Content-Transfer-Encoding: 8bit\n"
+      "POT-Creation-Date: 2018-05-02T22:06:24+00:00\n"
+      "PO-Revision-Date: 2018-05-02T22:06:24+00:00\n"
+      "X-Domain: foo-plugin\n"
+      "Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+      #: foo-plugin.js:15
+      msgid "Foo Plugin"
+      msgstr "Foo Plugin"
+      """
+
+    When I run `wp i18n make-php foo-plugin`
+    Then STDOUT should contain:
+      """
+      Success: Created 1 file.
+      """
+    And the return code should be 0
+    And the foo-plugin/foo-plugin-de_DE.php file should exist
+
+  Scenario: Allows setting custom destination directory
+    Given an empty foo-plugin directory
+    And a foo-plugin/foo-plugin-de_DE.po file:
+      """
+      # Copyright (C) 2018 Foo Plugin
+      # This file is distributed under the same license as the Foo Plugin package.
+      msgid ""
+      msgstr ""
+      "Project-Id-Version: Foo Plugin\n"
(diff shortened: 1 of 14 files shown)


danielbachhuber:
@swissspidy What sort of review would you like on this?

src/MakePhpCommand.php
```diff
@@ -0,0 +1,88 @@
...
+				continue;
+			}
+
+			if ( ! $file->isFile() || ! $file->isReadable() ) {
+				WP_CLI::warning( sprintf( 'Could not read file %s', $file->getFilename() ) );
+				continue;
+			}
+
+			$file_basename    = basename( $file->getFilename(), '.po' );
+			$destination_file = "{$destination}/{$file_basename}.php";
```
swissspidy, line 75: Note to self: the file suffix changed to `.mo.php`. Maybe further changes to `.l10n.php` or something else too.

swissspidy:
@danielbachhuber just a general review if the code and code style look sane. Functionality-/logic-wise I will be syncing this with any developments in core (see https://make.wordpress.org/core/2023/11/08/merging-performant-translations-into-core/ for the latest post)

src/MakePhpCommand.php
```diff
@@ -0,0 +1,88 @@
...
+	 * [<destination>]
+	 * : Path to the destination directory for the resulting PHP files. Defaults to the source directory.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     # Create PHP files for all PO files in the current directory.
+	 *     $ wp i18n make-php .
+	 *
+	 *     # Create a PHP file from a single PO file in a specific directory.
+	 *     $ wp i18n make-php example-plugin-de_DE.po languages
```
danielbachhuber, line 31: Should we show some sample output on these examples?

src/MakePhpCommand.php
```diff
@@ -0,0 +1,88 @@
...
+	 *
+	 * @when before_wp_load
+	 *
+	 * @throws WP_CLI\ExitException
+	 */
+	public function __invoke( $args, $assoc_args ) {
+		$source = realpath( $args[0] );
+		if ( ! $source || ( ! is_file( $source ) && ! is_dir( $source ) ) ) {
+			WP_CLI::error( 'Source file or directory does not exist!' );
```
danielbachhuber, line 40: I don't think the exclamation mark is necessary. We don't generally use such elsewhere.

src/MakePhpCommand.php
```diff
@@ -0,0 +1,88 @@
...
+		if ( isset( $args[1] ) ) {
+			$destination = $args[1];
+		}
+
+		// Two is_dir() checks in case of a race condition.
+		if ( ! is_dir( $destination )
+			&& ! mkdir( $destination, 0777, true )
+			&& ! is_dir( $destination )
+		) {
+			WP_CLI::error( 'Could not create destination directory!' );
```
danielbachhuber, line 53: Ditto above.

src/MakePhpCommand.php
```diff
@@ -0,0 +1,88 @@
...
+		if ( ! $source || ( ! is_file( $source ) && ! is_dir( $source ) ) ) {
+			WP_CLI::error( 'Source file or directory does not exist!' );
+		}
+
+		$destination = is_file( $source ) ? dirname( $source ) : $source;
+		if ( isset( $args[1] ) ) {
+			$destination = $args[1];
+		}
+
+		// Two is_dir() checks in case of a race condition.
```
danielbachhuber, line 48: What's the potential race condition?

src/MakePhpCommand.php
```diff
@@ -0,0 +1,88 @@
...
+			$translations = Translations::fromPoFile( $file->getPathname() );
+			if ( ! PhpArrayGenerator::toFile( $translations, $destination_file ) ) {
+				WP_CLI::warning( sprintf( 'Could not create file %s', $destination_file ) );
+				continue;
+			}
+
+			++$result_count;
+		}
+
+		WP_CLI::success( sprintf( 'Created %d %s.', $result_count, Utils\pluralize( 'file', $result_count ) ) );
```
danielbachhuber, line 86: Do we want to specify the type of file?

features/makephp.feature
```diff
@@ -0,0 +1,167 @@
...
+    When I run `wp i18n make-php foo-plugin`
+    Then STDOUT should contain:
+      """
+      Success: Created 1 file.
+      """
+    And the return code should be 0
+    And the foo-plugin/foo-plugin-de_DE.php file should contain:
+      """
+      'messages'=>[''=>['Foo Plugin'=>['Bar Plugin']]]
+      """
```
danielbachhuber, line 167: Should we have a test that verifies a valid PHP translation file is produced?

danielbachhuber (commented):
Command code generally seems fine. I didn't test the actual functionality, though. I left a few nits inline.

features/makephp.feature
swissspidy: Valid as in syntax or as in translation gets loaded properly in WordPress?

src/MakePhpCommand.php
swissspidy: We don‘t do that for any of the other commands I think.

src/MakePhpCommand.php
swissspidy: I think this goes back to https://github.com/kalessil/phpinspectionsea/blob/master/docs/probable-bugs.md#mkdir-race-condition

This particular code is also to be found in MakePotCommand.

Given that this isn‘t a concurrent system, we can probably just remove this in both places.

src/MakePhpCommand.php
swissspidy: Good point. I think I copied this from the other commabds so I‘ll make sure usage is aligned everywhere.

src/MakePhpCommand.php
swissspidy: Command ouput or actual file contents?

features/makephp.feature
danielbachhuber: I was thinking the latter but we could check both, I suppose.

src/MakePhpCommand.php
danielbachhuber: Command output.

features/makephp.feature
swissspidy: This is probably best done in a future PR because we don't have similar assertions for the other file formats, and the PHP format especially is not yet merged into core.

danielbachhuber: Sounds good 👍

danielbachhuber (approved):

Issue #176, opened 2019-07-24
Support for TypeScript .tsx files


matzeeable:
## Feature Request

- [X] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

When using `wp i18n make-pot` within a source folder containing only `.tsx` TypeScript files, no gettext translations are extracted. Using the command on the compiled source files `.js` and `.map` files it works correctly - but slow.

**Describe the solution you'd like**

I have worked a long time on a real and complete boilerplate plugin https://github.com/matzeeable/wp-reactjs-starter. It also supports generation of `.pot` and `.json` files for frontend. But it is completely written in TypeScript for frontend. So, is there a way to extract the gettext functions from `.tsx` files?

Would it be enough to add those extensions to this call:

https://github.com/wp-cli/i18n-command/blob/af8143381e6a96555b533471e8f9680d755d42d9/src/MakePotCommand.php#L572-L580

Reference: matzeeable/wp-reactjs-starter#18

swissspidy:
> Would it be enough to add those extensions to this call:

No, because we don't have a parser that supports TypeScript. We currently use [Peast](https://github.com/mck89/peast), which supports JavaScript and JSX only.

If we'd want to support TypeScript, we need to be able to parse it.

Perhaps you can ask the Peast developer whether it's on their roadmap?

matzeeable:
Thanks for your fast response. Ok I will further look into this soon. On the other hand I think parsing TypeScript can be a very huge and time-intensive programming task. Perhaps there is a more simpler way doing this, e. g. transforming TSX to plain JS temporarely with the `ts-loader` within WebPack, i18n' it and then bundle it together. Close... :-)

vanduc1102:
currently, I use the result of `yarn build` --> `main.js` ( without minimized ) to generate  pot content,
so you have to run `yarn build` then `wp i18n make-pot`
it is working fine,  only small  issue that comment for the message is referring to `main.js` , for my project, it is acceptable.

tyrann0us:
@swissspidy

> Perhaps you can ask the Peast developer whether it's on their roadmap?

Done in https://github.com/mck89/peast/issues/28#issuecomment-1302106000.

tyrann0us:
Hi @swissspidy, admittedly, I don’t know anything about language parsers and ASTs.
[...]

PR #364, opened 2023-06-01
make-pot: Add TypeScript support

retlehs:
This PR updates the `make-pot` command to support scanning files that end with `.ts` and `.tsx` to support TypeScript

### DIFF
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -657,7 +657,7 @@ class MakePotCommand extends WP_CLI_Command {
 					[
 						'include'       => $this->include,
 						'exclude'       => $this->exclude,
-						'extensions'    => [ 'js', 'jsx' ],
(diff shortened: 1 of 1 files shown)


swissspidy:
Thanks for raising this, but unfortunately it doesn‘t work like that. To support TypeScript files we need a parser that supports TS, and ours doesn‘t. And I‘m not aware of another JS/TS parser in PHP, especially since there is no proper TS spec anyway that could be implemented.

Adding TS support not only means having such a parser, but also tests as well.

See #176 where this was suggested before.

Out of curiosity, why do you need this? The make-pot command has been developed for using it on built JS files that run in the browser

retlehs:
> The make-pot command has been developed for using it on built JS files that run in the browser

This makes sense :+1: I'm used to running the make-pot command on a directory with pre-built JS files, but it's not a problem to run it on the compiled files instead

PR #366, opened 2023-06-28
Update `theme-i18n.json`

swissspidy:
~We really need a better way to keep this file in sync with core.~ Edit: forgot that we fetch directly from svn

### DIFF
--- a/.github/workflows/update-schemas.yml
+++ b/.github/workflows/update-schemas.yml
@@ -1,0 +1,76 @@
+name: Update i18n schemas
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
+    paths-ignore:
+      - "features/**"
+      - "README.md"
+
+
+# Cancels all previous workflow runs for the same branch that have not yet completed.
+concurrency:
+  # The concurrency group contains the workflow name and the branch name.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+
+  update-schemas: #----------------------------------------------------------
+    name: Update i18n schema fallback files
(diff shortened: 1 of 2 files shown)


oandregal:
I've compared the changes with core and it looks good. I'm trying to find the tests instructions in older PRs.

oandregal:
This is what I've done for testing (updated from https://github.com/wp-cli/i18n-command/pull/306):

- `composer install`
- Update the `THEME_JSON_SOURCE` to be empty in [this line](https://github.com/wp-cli/i18n-command/blob/3a18b5e5f3a1d231938da4669c1ea629e063f68a/src/JsonSchemaExtractor.php#L18) at the `JsonSchemaExtractor` class, so the command uses the local backup file defined by `THEME_JSON_FALLBACK`.
- Create a new directory called `foo-theme` that contains a `theme.json` file and paste the content of [TwentyTwentyThree's theme.json](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-content/themes/twentytwentythree/theme.json).
- Verify that there's a `foo-theme/foo-theme.pot` file and that strings and their context are present (look for "Custom template name", "Template part name", "Font family name", "Space size name", the ones added in this PR).

swissspidy:
Awesome, thanks for testing! 🎉 

Totally forgot that this is just the backup file. That makes the change way less urgent 🙂 

I'll see if I can reproduce that test failure locally.

schlessera:
> We really need a better way to keep this file in sync with core.
We can add a GHA that replaces the included file with the one on SVN and creates a PR if changes have been detected.

swissspidy:
@schlessera Good idea. Just added such a workflow now.

danielbachhuber (approved):

Issue #372, opened 2023-11-08
`make-mo`: Add option to specify filename


marekdedic:
## Feature Request

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

I'm localizing a WordPress theme that unfortunately cannot be hosted on WP.org due to organisation policy. I am trying to migrate from a collection of tools to the i18n command. I need to create `.mo` file to be loaded with `load_theme_textdomain()`. My `.po` files are called like `theme-slug-cs_CZ.po` (so theme slug + locale name), however, the `load_theme_textdomain()` function requires the `.mo` file to be called just `cs_CZ.mo` (so just locale name). This requires me to strip the theme slug from all `.mo` files

**Describe the solution you'd like**

I'd like to see the option to use just the locale name for the `make-mo` command

swissspidy:
I‘d be open for a PR that provides an option to override the name.

But why not just rename the PO file?

marekdedic:
Yeah, a rename would solve the issue, true. I thought the convention was to include the plugin/theme name, though - I don't know why I think that, I found e.g. https://developer.wordpress.org/plugins/internationalization/localization/#translate-the-po-file now...

swissspidy:
If you provide your own translation files, the file name convention is `<slug>-<locale>.mo` for _plugins_ and `<locale>.mo` for _themes_. The difference is mostly because of historical reasons.

marekdedic:
Thanks, didn't know that. Then I'll solve my immediate issue by renaming the po files, however, I still think a (probably more general) option to set the file names could be useful...

sdnunca:
@swissspidy I've looked into adding the option of overriding the file name, however given that the command also supports converting all files from a directory and in that case the name override would not make much sense. 

I'm not sure if adding an option to override the filename that only works when you pass file would make much sense, would appreciate any feedback.

swissspidy:
That's a good point. I suppose it could still work though, just like [`msgfmt`](https://www.gnu.org/software/gettext/manual/html_node/msgfmt-Invocation.html) supports both a directory and a single file as input, and supports a `--output-file` argument.

Here, we could just enhance `wp i18n make-mo <source> [<destination>]` so that if `source` is a single file and `destination` is a file, then we would use that file name instead.

Example:

[...]

PR #373, opened 2023-11-16
`make-mo`: Add destination file support

sdnunca:
Based on the discussion in #372, the `MakeMo `command should support a destination file name.

This PR allows you to specify a destination file name when the source is a file. When the source is a directory, providing a destination file name is not supported:

`wp i18n make-mo foo.po` -> unchanged behavior, writes foo.mo to same directory as foo.po
`wp i18n make-mo foo.po bar.mo` -> writes file to bar.mo
`wp i18n make-mo foo/bar` -> unchanged behavior, generates MO files for all PO files in the directory
`wp i18n make-mo foo/bar baz.mo` -> won't work!

### DIFF
--- a/src/MakeMoCommand.php
+++ b/src/MakeMoCommand.php
@@ -20,7 +20,7 @@ class MakeMoCommand extends WP_CLI_Command {
 	 * : Path to an existing PO file or a directory containing multiple PO files.
 	 *
 	 * [<destination>]
-	 * : Path to the destination directory for the resulting MO files. Defaults to the source directory.
+	 * : Path to the destination file or directory for the resulting MO files. Defaults to the source directory.
 	 *
 	 * ## EXAMPLES
 	 *
@@ -29,6 +29,9 @@ class MakeMoCommand extends WP_CLI_Command {
 	 *
 	 *     # Create a MO file from a single PO file in a specific directory.
 	 *     $ wp i18n make-mo example-plugin-de_DE.po languages
+	 *
+	 *     # Create a MO file from a single PO file to a specific file destination
+	 *     $ wp i18n make-mo example-plugin-de_DE.po languages/bar.mo
 	 *
 	 * @when before_wp_load
 	 *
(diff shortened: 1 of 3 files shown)


tests/MakeMoTest.php
```diff
@@ -0,0 +1,74 @@
...
+		$make_mo = new MakeMoCommand();
+		$make_mo( array(
+			$base_directory,
+		), array() );
+		$this->assertTrue( is_file( $base_directory . '/test.mo' ), 'MO file not generated' );
+		unlink( $file_path );
+		unlink( $base_directory . '/test.mo' );
+		rmdir( $base_directory );
+	}
+}
```
danielbachhuber, line 74: Sorry for the confusion!

Can you convert these tests to Behat tests? While they look like perfectly valid PHPUnit tests, we use Behat for our functional tests. Here's a primer: https://make.wordpress.org/cli/handbook/contributions/pull-requests/#functional-tests

src/MakeMoCommand.php
```diff
@@ -40,9 +43,19 @@ public function __invoke( $args, $assoc_args ) {
 			WP_CLI::error( 'Source file or directory does not exist!' );
 		}
 
-		$destination = is_file( $source ) ? dirname( $source ) : $source;
+		$destination      = is_file( $source ) ? dirname( $source ) : $source;
+		$custom_file_name = null;
 		if ( isset( $args[1] ) ) {
-			$destination = $args[1];
+			$destination           = $args[1];
+			$destionation_pathinfo = pathinfo( $destination );
```
swissspidy, line 50: Typo

```suggestion
			$destination_pathinfo = pathinfo( $destination );
```

tests/MakeMoTest.php
sdnunca: Thank you, it seems I missed them. Updated.

src/MakeMoCommand.php
sdnunca: Thank you, updated.

danielbachhuber:
Looks good to me! I'll let @swissspidy give it a final review.

swissspidy (approved):

Issue #65, opened 2018-07-31
Use extractor instead of regex to get main file header data


swissspidy:
This is something I noticed while working on #61 / #64.

When printing warnings for missing / invalid strings I want to print the warning together with the location of where that string is.

For file headers that's currently not possible because they're being extracted using regex. This is mainly because I borrowed that code from https://develop.svn.wordpress.org/trunk/tools/i18n/makepot.php, which has been doing that for years.

However, since we already leverage a powerful gettext extraction library, I wonder if we could add a new extractor just for these file headers.

`retrieve_main_file_data()`, `get_file_data()` `get_file_headers()`, `get_file_data_from_string()`, and `_cleanup_header_comment()` would be affected/obsoleted by such a change.

I'm not sure how feasible it is, but it would make for a more robust solution that is more similar to the other extractors and easier to test than some old regex from WordPress core.

Also, we could make sure to get the filename / line references that way.

I guess it's the same argument as regex vs. using a DOM parser to parse some HTML in PHP.

----

I just saw that the `Template Name:` extraction already resides in `\WP_CLI\I18n\IterableCodeExtractor`.

We should probably consolidate this somehow.

swissspidy:
Another benefit:

For WordPress core we wouldn't need to extract Hello Dolly separately, but simultaneously _and_ with the correct location info.

We'd still need some way to detect whether the project is a plugin, theme, core, or something else.

swissspidy:
Follow-up: #376

swissspidy:
See https://github.com/wp-cli/i18n-command/pull/377#issuecomment-1861060828 for how this could maybe be achieved with regex.

PR #377, opened 2023-12-18
Add file references for plugin/theme headers

### LINKED ISSUE #376: Consider adding file reference for strings extracted from file headers
## Feature Request

Context:

* https://github.com/swissspidy/preferred-languages/issues/883
* https://meta.trac.wordpress.org/ticket/7376

Since there is no file reference for strings extracted from plugin/theme headers, it's easy to miss where they're from.

This was originally not added to keep feature parity with the old make-pot tool, but it seems reasonable to add.

Add source file reference for things like plugin descriptions to avoid issues like the above.

swissspidy:
Fixes #376

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -66,6 +66,40 @@ Feature: Generate a POT file of a WordPress project
       Description of the plugin
       """
 
+  Scenario: Adds file references for file headers.
+    Given an empty foo-plugin directory
+    And a foo-plugin/foo-plugin.php file:
+      """
+      <?php
+      /**
+       * Plugin Name: Foo Plugin
+       * Plugin URI:  https://example.com
+       * Description: Plugin Description
+       * Version:     0.1.0
+       * Author:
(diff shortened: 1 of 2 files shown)


swissspidy:
FYI @ocean90 — would like to know if you think this makes sense or not.

ocean90 (approved):
LGTM.

Would be nice to have the line numbers as well but not a blocker for me.

swissspidy:
Yeah I thought about that as well. Always thought that requires a more complex solution (see #65) but maybe using `PREG_OFFSET_CAPTURE` would suffice (see [this SO post](https://stackoverflow.com/questions/4730867/get-line-number-from-preg-match-all)). I'll see how feasible it is and otherwise will file a follow-up issue.

swissspidy:
@wp-cli/committers would appreciate a quick review here so I can merge this

danielbachhuber (approved):

PR #384, opened 2024-02-13
Use relative paths for file header references

### LINKED ISSUE #383: Are file references for plugin/theme headers supposed to be absolute paths?
I realize support questions are supposed to go through other channels, but since this is closely related to #377 and #65, posting here made sense.  (Though, let me know if elsewhere would still be better).  

I'm a plugin developer and I'm noticing the header strings now being added to pot files, which is great!  But, the file references being used are absolute paths, which seems odd to me.  

I'm running on Windows, but this is holding true for the handful of plugins I work on. 

Running `wp-cli i18n make-pot . i18n/my-plugin.pot` from the root directory of the plugin results in the pot file containing:

```
...
#. Author URI of the plugin
#: V:\my-plugin\my-plugin.php
msgid "https://github.com/jkrrv"
msgstr ""

#: src/templates/admin/invKoForm.php:17
#: src/templates/admin/locationsKoForm.php:13
#: src/templates/admin/locationsKoForm.php:50
msgid "Delete"
msgstr ""
...
```

I've included the last header item, and the first regular string.  The second item, the regular string, has the relative paths I expect.  But the paths for the header strings are all absolute, which is not what I would expect.  After all, these paths are going to be different for every developer or translator. 

Is this intentional?  

Thanks!

jkrrv:
Resolves #383

### DIFF
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -628,7 +628,7 @@ class MakePotCommand extends WP_CLI_Command {
 
 			if ( $this->main_file_path && $this->location ) {
 				$translation->addReference(
-					str_replace( "$this->source/", '', $this->main_file_path )
+					str_replace( "$this->source/", '', wp_normalize_path( $this->main_file_path ) )
 				);
 			}
 
--- a/src/MakePotCommand.php
+++ b/src/MakePotCommand.php
@@ -628,7 +628,7 @@ class MakePotCommand extends WP_CLI_Command {
 
 			if ( $this->main_file_path && $this->location ) {
 				$translation->addReference(
-					str_replace( "$this->source/", '', wp_normalize_path( $this->main_file_path ) )
+					str_replace( "$this->source/", '', Utils\normalize_path( $this->main_file_path ) )
 				);
 			}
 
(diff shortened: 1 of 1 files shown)


swissspidy:
Thanks for the fix!

Not sure why the test didn't catch this:

https://github.com/wp-cli/i18n-command/blob/ca7a44a1f8b09d533621b4006e739974212860ee/features/makepot.feature#L69-L101

src/MakePotCommand.php
```diff
@@ -628,7 +628,7 @@ protected function extract_strings() {
 
 			if ( $this->main_file_path && $this->location ) {
 				$translation->addReference(
-					str_replace( "$this->source/", '', $this->main_file_path )
+					str_replace( "$this->source/", '', Utils\normalize_path( $this->main_file_path ) )
```
swissspidy, line 631: In other places we use `ltrim` as well:

```suggestion
					ltrim( str_replace( "$this->source/", '', Utils\normalize_path( $this->main_file_path ) ), '/' )
```

jkrrv:
Awesome.  Thanks for the speedy response.

swissspidy (approved):

Issue #395, opened 2024-04-09
make-php - Make output file content pretty


sovetski:
## Feature Request

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

When we do `wp i18n make-php` it creates PHP files containing one line, like minified PHP, but If we want to delete the old .po/.mo files to work with only PHP translations, working one line generated files is not practical

![image](https://github.com/wp-cli/i18n-command/assets/13520683/a933ee0b-3f56-4387-bd66-f9fc960d4933)

**Describe the solution you'd like**

Formatted output will look better

![image](https://github.com/wp-cli/i18n-command/assets/13520683/4f5318dc-6ec8-4a22-bc17-f6d8ba5c5051)

swissspidy:
We can add a `pretty-print` option like we have for make-json. Wanna help put a PR together? :)

sovetski:
> We can add a `pretty-print` option like we have for make-json. Wanna help put a PR together? :)

Done: https://github.com/wp-cli/i18n-command/pull/396

Let me know if something is not OK 😊

sovetski:
@kkmuffme yes exactly it is to not use .po files anymore, php files are more readable and easy to edit

PR #396, opened 2024-04-09
Add support for pretty-printing PHP files in wp i18n make-php command

sovetski:
It should resolve this issue: https://github.com/wp-cli/i18n-command/issues/395

### DIFF
--- a/features/makephp.feature
+++ b/features/makephp.feature
@@ -259,3 +259,48 @@ Feature: Generate PHP files from PO files
       """
       new message
       """
+
+  Scenario: Should create pretty-printed PHP files
+    Given an empty foo-plugin directory
+    And a foo-plugin/foo-plugin-de_DE.po file:
+      """
+      # Copyright (C) 2018 Foo Plugin
+      # This file is distributed under the same license as the Foo Plugin package.
+      msgid ""
+      msgstr ""
+      "Project-Id-Version: Foo Plugin\n"
+      "Report-Msgid-Bugs-To: https://wordpress.org/support/plugin/foo-plugin\n"
+      "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+      "Language-Team: LANGUAGE <LL@li.org>\n"
+      "Language: de_DE\n"
+      "MIME-Version: 1.0\n"
+      "Content-Type: text/plain; charset=UTF-8\n"
+      "Content-Transfer-Encoding: 8bit\n"
+      "POT-Creation-Date: 2018-05-02T22:06:24+00:00\n"
+      "PO-Revision-Date: 2018-05-02T22:06:24+00:00\n"
+      "X-Domain: foo-plugin\n"
+      "Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+      #: foo-plugin.js:15
+      msgid "Foo Plugin"
+      msgstr "Foo Plugin"
+      """
+
+    When I run `wp i18n make-php foo-plugin`
+    Then STDOUT should contain:
+      """
+      Success: Created 1 file.
+      """
+    And the return code should be 0
+    And the foo-plugin/foo-plugin-de_DE.php file should contain:
+      """
+      <?php
+      return [
+          'project-id-version' => 'Foo Plugin',
+          'report-msgid-bugs-to' => 'https://wordpress.org/support/plugin/foo-plugin',
+          'messages' =>
+              [
+                  'Foo Plugin' => 'Foo Plugin',
+              ],
+      ];
+      """
--- a/src/PhpArrayGenerator.php
+++ b/src/PhpArrayGenerator.php
@@ -14,6 +14,7 @@ use Gettext\Translations;
 class PhpArrayGenerator extends PhpArray {
 	public static $options = [
 		'includeHeaders' => false,
+		'prettyPrint'    => false,
 	];
 
 	/**
@@ -22,7 +23,10 @@ class PhpArrayGenerator extends PhpArray {
 	public static function toString( Translations $translations, array $options = [] ) {
 		$array = static::generate( $translations, $options );
 
-		return '<?php' . PHP_EOL . 'return ' . static::var_export( $array ) . ';';
+		$pretty_print   = isset( $options['prettyPrint'] ) ? $options['prettyPrint'] : false;
+		$exported_array = static::var_export( $array, $pretty_print );
+
+		return '<?php' . PHP_EOL . 'return ' . $exported_array . ';';
 	}
 
 	/**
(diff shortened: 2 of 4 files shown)


features/makephp.feature
```diff
@@ -259,3 +259,48 @@ Feature: Generate PHP files from PO files
...
+      msgstr "Foo Plugin"
+      """
+
+    When I run `wp i18n make-php foo-plugin`
+    Then STDOUT should contain:
+      """
+      Success: Created 1 file.
+      """
+    And the return code should be 0
+    And the foo-plugin/foo-plugin-de_DE.php file should contain:
```
swissspidy, line 295: The file extension is .l10n.php

features/makephp.feature
```diff
@@ -259,3 +259,48 @@ Feature: Generate PHP files from PO files
...
+      <?php
+      return [
+          'project-id-version' => 'Foo Plugin',
+          'report-msgid-bugs-to' => 'https://wordpress.org/support/plugin/foo-plugin',
+          'messages' =>
+              [
+                  'Foo Plugin' => 'Foo Plugin',
+              ],
+      ];
+      """
```
swissspidy, line 306: Nit: missing newline at end of file

features/makephp.feature
```diff
@@ -259,3 +259,48 @@ Feature: Generate PHP files from PO files
...
+      "Language: de_DE\n"
+      "MIME-Version: 1.0\n"
+      "Content-Type: text/plain; charset=UTF-8\n"
+      "Content-Transfer-Encoding: 8bit\n"
+      "POT-Creation-Date: 2018-05-02T22:06:24+00:00\n"
+      "PO-Revision-Date: 2018-05-02T22:06:24+00:00\n"
+      "X-Domain: foo-plugin\n"
+      "Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+      #: foo-plugin.js:15
```
swissspidy, line 284: Maybe test with 1-2 more entries too, perhaps also a plural one

features/makephp.feature
```diff
@@ -259,3 +259,61 @@ Feature: Generate PHP files from PO files
...
+
+    When I run `wp i18n make-php foo-plugin`
+    Then STDOUT should contain:
+      """
+      Success: Created 1 file.
+      """
+    And the return code should be 0
+    And the foo-plugin/foo-plugin-de_DE.l10n.php file should contain:
+      """
+      <?php
```
ernilambar, line 307: Should not this be the output when `--pretty-print` is passed in `wp i18n make-php foo-plugin`?

ernilambar:
@sovetski FYI: You can run Behat test locally easily using commands like: 
```
WP_CLI_TEST_DBTYPE=sqlite composer behat features/makephp.feature:263
```
This will run behat test of given file and 263 is the line number of your new Scenario. This will be helpful for quick test in local environment. With using `sqlite`, you don't have to face hassle to setup MySQL.

features/makephp.feature
swissspidy: Not sure what you mean — we always need an opening PHP tag.

ernilambar: `wp i18n make-php foo-plugin --pretty-print` should give this output but in line 299, `wp i18n make-php foo-plugin` is kept.

swissspidy: Ah, I see. Indeed, the test should actually be using `--pretty-print`

features/makephp.feature
```diff
@@ -259,3 +259,61 @@ Feature: Generate PHP files from PO files
...
+      msgstr "Hallo"
+
+      #: foo-plugin.js:17
+      msgid "You have %d new message"
+      msgid_plural "You have %d new messages"
+      msgstr[0] "Du hast %d neue Nachricht"
+      msgstr[1] "Du hast %d neue Nachrichten"
+      """
+
+    When I run `wp i18n make-php foo-plugin`
```
swissspidy, line 299:
```suggestion
    When I run `wp i18n make-php foo-plugin --pretty-print`
```

sovetski:
> @sovetski FYI: You can run Behat test locally easily using commands like:
> 
> ```
> WP_CLI_TEST_DBTYPE=sqlite composer behat features/makephp.feature:263
> ```
> 
> This will run behat test of given file and 263 is the line number of your new Scenario. This will be helpful for quick test in local environment. With using `sqlite`, you don't have to face hassle for MySQL setup.

Thank you, it helped a lot, now the test pass successfully

image

src/PhpArrayGenerator.php
```diff
@@ -163,4 +172,36 @@ private static function var_export( $value ) {
...
 		return '[' . implode( ',', $entries ) . ']';
 	}
+
+	/**
+	 * Outputs or returns a parsable string representation of a variable.
+	 * @since 4.0.0
+	 *
+	 * @param mixed $value The variable you want to export.
+	 * @param int   $level The current indentation level.
+	 * @param int   $indentation The number of spaces for indentation. Default is 4.
```
swissspidy, line 182: Can we do tabs instead?

sovetski:
I don't understand. The command `WP_CLI_TEST_DBTYPE=sqlite composer behat features/makephp.feature:263` was working as well, I just removed 1 space and adapted my scenario, I have this error without any explanation:

image

When I add `-vvv` it returns an unreadable PHP exception. But the returned error output matches the scenario's expected output, so very weird...

src/PhpArrayGenerator.php
sovetski: I like using tabs for my own projects, but here as it is something which will be used on other developers IDE, I prefer to use spaces, because tabs configuration depends on the configuration of user. A space will be every time a space, but tab not. Some resources not related to this subject but still on the same issue:

https://medium.com/@peey/what-is-the-inconsistent-use-of-tabs-and-spaces-in-indentation-error-and-why-is-it-caused-f3bbb8b2568b
https://stackoverflow.com/questions/5685406/inconsistent-use-of-tabs-and-spaces-in-indentation

swissspidy: > I prefer to use spaces, because tabs configuration depends on the configuration of user

That's actually an advantage of tabs :-) But I don't wanna start yet another tabs vs spaces discussion here.

I am mainly suggesting using tabs because that's the WordPress coding standard and IMHO we should be consistent with that.

sovetski: Great, in this case I will try to change it to tabs to respect the standards, thanks!

swissspidy:
> When I add `-vvv` it returns an unreadable PHP exception. But the returned error output matches the scenario's expected output, so very weird...

It sounds like it doesn't match exactly.

When I manually run the `wp i18n` command in my terminal, I get this output with lots of indentation (8 spaces):

```
<?php
return [
        'domain' => 'foo-plugin',
        'plural-forms' => 'nplurals=2; plural=(n != 1);',
        'language' => 'de_DE',
        'project-id-version' => 'Foo Plugin',
        'pot-creation-date' => '2018-05-02T22:06:24+00:00',
        'po-revision-date' => '2018-05-02T22:06:24+00:00',
        'messages' => [
                'Foo Plugin' => 'Foo Plugin',
                'Hello' => 'Hallo',
                'You have %d new message' => 'Du hast %d neue Nachricht' . "\0" . 'Du hast %d neue Nachrichten',
        ],
];
````

Issue #398, opened 2024-04-26
Incorrect POT license info for themes without a license in the header


meszarosrob:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

If in a Theme Header (`style.css`) the `License` is *not* added,  when I generate the POT file, I see the `This file is distributed under the .`

https://github.com/wp-cli/i18n-command/blob/e29d18189b4291eca9d06b7da1e85d4736d92c5a/src/MakePotCommand.php#L907

**Describe how other contributors can replicate this bug**

```
Scenario: Wrong license information is displayed for the theme
  Given an empty foo-theme directory
  And a foo-theme/style.css file:
    """
    /*
    Theme Name: Foo Theme
    */
    """

  When I run `wp i18n make-pot foo-theme foo-theme.pot`
  Then the foo-theme.pot file should contain:
    """
    # This file is distributed under the .
    """
```

**Describe what you would expect as the correct outcome**

To see the `This file is distributed under the same license as the X theme.`

https://github.com/wp-cli/i18n-command/blob/e29d18189b4291eca9d06b7da1e85d4736d92c5a/src/MakePotCommand.php#L915

**Let us know what environment you are running this on**

```
OS:     Linux 6.5.0-10027-tuxedo
Shell:  
PHP binary:     /usr/bin/php8.0
PHP version:    8.0.30
php.ini used:   /etc/php/8.0/cli/php.ini
MySQL binary:   /usr/bin/mysql
MySQL version:  mysql  Ver 15.1 Distrib 10.5.21-MariaDB, for debian-linux-gnu (x86_64) using  EditLine wrapper
SQL modes:      STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION
WP-CLI root dir:        phar://wp-cli.phar/vendor/wp-cli/wp-cli
WP-CLI vendor dir:      phar://wp-cli.phar/vendor
WP_CLI phar path:       /var/www/html
WP-CLI packages dir:    
WP-CLI cache dir:       /home/user/.wp-cli/cache
WP-CLI global config:   
[...]

PR #399, opened 2024-04-26
Ensure that the POT file use the same license as the theme

meszarosrob:
Fixes https://github.com/wp-cli/i18n-command/issues/398


swissspidy:
Thanks for your contribution during the [WP-CLI Hack day](https://github.com/wp-cli/wp-cli/issues/5935)!

swissspidy (approved):

PR #400, opened 2024-05-14
Use relative paths for file header references (Windows fix)

saas786:
Ref:

* https://github.com/wp-cli/i18n-command/pull/384
* https://github.com/wp-cli/i18n-command/issues/383


swissspidy (approved):
Thank you!

PR #404, opened 2024-06-13
Add tests for rspack and esbuild bundle output

matzeeable:
This is a follow-up to https://github.com/wp-cli/i18n-command/issues/203 and https://github.com/wp-cli/i18n-command/pull/204. 

In general, the `wp i18n make-pot` already works with [rspack](https://www.rspack.dev/) and [esbuild](https://esbuild.github.io/getting-started/#your-first-bundle) and can extract the strings correctly from the generated bundle.

This PR just adds test for both bundlers.

Reference. https://github.com/wp-cli/wp-cli/issues/5955


swissspidy (approved):
Thank you!

PR #424, opened 2024-12-19
make-pot: scan any `theme.json` file in any level

### LINKED ISSUE #423: Fix `theme.json` string extraction
In #320 we limited extraction of `theme.json` files to ones in the (theme) root and in a `/styles` folder.

We forgot the fact that there's `wp-includes/theme.json` in core and `lib/theme.json` in Gutenberg which are currently forgotten because of that, thus those strings aren't extracted nor translatable in these projects. But they should be!

We probably need to drop that limitation and just scan _any_ `theme.json` file we encounter...

Meta changes (in [`sites/trunk/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-make-core-pot.php`](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-make-core-pot.php)) are probably required too once that lands. To avoid any regression there.

/cc @ocean90 @oandregal

swissspidy:
* scan any `theme.json` file in any level, for any type of project
* only scan style variations in a top-level `/styles` folder **if it is a theme** (this restriction is new)

Fixes #423


swissspidy:
@oandregal @ocean90 would appreciate your feedback on this.

oandregal:
I've looked at [the command](https://github.com/WordPress/wordpress.org/blob/83888a04df5f3d7d218503755543a73a6235c3ef/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-make-core-pot.php#L261) used to extract core strings. It creates a few different files:

- [wordpress-continents-cities.pot](https://github.com/WordPress/wordpress.org/blob/cc33c6d00d8180c6eaa4fe67340c8b665faa3c46/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-make-core-pot.php#L67-L85): no strings coming from theme.json are related to this.
- [wordpress.pot](https://github.com/WordPress/wordpress.org/blob/83888a04df5f3d7d218503755543a73a6235c3ef/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-make-core-pot.php#L138-L158): this looks like the strings for the front-end? No strings coming from `theme.json` need to be displayed in the front-end.
- [hello-dolly.pot](https://github.com/WordPress/wordpress.org/blob/83888a04df5f3d7d218503755543a73a6235c3ef/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-make-core-pot.php#L138-L158): no strings coming from `theme.json` are related to the hello-dolly plugin.
- [wordpress-admin.pot](https://github.com/WordPress/wordpress.org/blob/83888a04df5f3d7d218503755543a73a6235c3ef/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-make-core-pot.php#L212-L236): strings to load on the admin. The strings coming from theme.json need to be available in all block editors. If this is the file we load for that, they need to be loaded here.
- [wordpress-admin-network.pot ](https://github.com/WordPress/wordpress.org/blob/83888a04df5f3d7d218503755543a73a6235c3ef/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-make-core-pot.php#L240-L261): this sounds like the admin network screen. No strings coming from theme.json here.

I suppose we should exclude `wp-content` from all those contexts, apart for some minor tweaks (include dolly, etc.). This would suffice to prevent theme-defined strings (be in PHP like patterns or functions.php, or any of the theme.json (main or  style variations) from being picked for the core pot.

With the limited context I have about how the extraction works, I believe the i18n-command should never skip a path. It's the consumer responsibility to configure the command properly and they can already `--exclude-*` paths and `--skip-*` types of files (PHP, JSON, etc.).

src/MakePotCommand.php
```diff
@@ -704,20 +704,38 @@ protected function extract_strings() {
...
 					$this->source,
 					$translations,
 					[
-						'schema'         => JsonSchemaExtractor::THEME_JSON_SOURCE,
-						'schemaFallback' => JsonSchemaExtractor::THEME_JSON_FALLBACK,
-						'include'        => $this->include,
-						'exclude'        => $this->exclude,
-						'extensions'     => [ 'json' ],
-						'addReferences'  => $this->location,
+						// Only look for theme.json files in any folder, nothing else.
```
oandregal, line 711: This means any file named `theme.json` at any level would be considered. This targets core's `wp-includes/theme.json`, Gutenberg's `theme.json`, and the top-level `theme.json`.

swissspidy: Correct

src/MakePotCommand.php
```diff
@@ -704,20 +704,38 @@ protected function extract_strings() {
...
+						'include'           => $this->include,
+						'exclude'           => $this->exclude,
+						'extensions'        => [ 'json' ],
+						'addReferences'     => $this->location,
 					]
 				);
+
+				// Themes can have style variations in the top-level "styles" folder.
+				// They're like theme.json but can have any name.
+				if ( $is_theme ) {
```
oandregal, line 724: And this targets theme style variations: any `*.json` file within the `styles` folder of a theme.

src/MakePotCommand.php
```diff
@@ -704,20 +704,38 @@ protected function extract_strings() {
...
+					JsonSchemaExtractor::fromDirectory(
+						$this->source,
+						$translations,
+						[
+							'restrictDirectories' => [ 'styles' ],
```
oandregal, line 729: The style variations within the `styles` folder can be organized in sub-folders ([example](https://github.com/WordPress/wordpress-develop/tree/trunk/src/wp-content/themes/twentytwentyfive/styles)). I presume this is already considered by this argument, but I haven't found it in the i18n-command codebase, so I can't read how it works. I wanted to raise this to make sure it works properly.

swissspidy: I'll add a test for this 👍

swissspidy: Test passes just fine

src/IterableCodeExtractor.php
```diff
@@ -38,8 +39,18 @@ public static function fromFile( $file_or_files, Translations $translations, arr
...
 			}
 
+			$relative_file_path = ltrim( str_replace( static::$dir, '', Utils\normalize_path( $file ) ), '/' );
+
 			// Make sure a relative file path is added as a comment.
-			$options['file'] = ltrim( str_replace( static::$dir, '', Utils\normalize_path( $file ) ), '/' );
+			$options['file'] = $relative_file_path;
+
+			if ( ! empty( $options['restrictDirectories'] ) ) {
+				$top_level_dirname = explode( '/', $relative_file_path )[0];
```
oandregal, line 48: This only considers the top-level `styles` folder of a theme — provided the root directory for the command is the theme folder. Is there any kind of support for themes that are nested within each other? Example: `wp-content/themes/my-theme-generator/my-child/`. There can be theme.json or style variations both `my-theme-generator` and `my-theme-generator/my-child`. I suppose the theme directory only admits "single themes", not "nested ones", so this wouldn't be an issue for string extraction.

swissspidy: > Is there any kind of support for themes that are nested within each other? Example: `wp-content/themes/my-theme-generator/my-child/`

Is that even a thing? 🤔 

If it's a nested theme that can be activated on its own, then it needs its own text domain anyway, so you wouldextract the strings for `my-child` individually, with `my-child` being the root.

> I suppose the theme directory only admits "single themes", not "nested ones", so this wouldn't be an issue for string extraction.

Correct.

oandregal (approved):
This direction sounds good to me, thanks for the quick fix.

I've shared an [edge case](https://github.com/wp-cli/i18n-command/pull/424#discussion_r1893769684) that we may want to address before merging.

src/MakePotCommand.php
swissspidy: Hmm interesting edge case.

I don't see a need to change anything about the processing though. If `styles/theme.json` ends up being processed twice it will only be in the POT file once. There won't be any duplicates whatsoever.

oandregal: Nice, thanks for confirming 👍

swissspidy:
Thanks @oandregal for your help and thinking of all possible edge cases here 🦸 🚀 

@wp-cli/committers if someone could give this a quick review & approval, that would be greatly appreciated!

ernilambar (approved):

PR #438, opened 2025-04-11
Fixed inaccurate make-mo command arguments docs.

sun:
Follow-up to
- #372 
- #373 

## Problem
The documentation on https://developer.wordpress.org/cli/commands/i18n/make-mo/ and in the help is inaccurate and misleading.  It states:
> ```
> # Create a MO file from a single PO file in a specific directory.
> $ wp i18n make-mo example-plugin-de_DE.po languages
> ```

But the suggested command yields an error:

```console
$ wp i18n make-mo woocommerce-de_DE.po wp-content/languages/plugins
Error: Source file or directory does not exist.
```

The correct command argument is to just specify the file including path:

```console
$ wp i18n make-mo wp-content/languages/plugins/woocommerce-de_DE.po
Success: Created 1 file.
```

## Proposed solution
Correct the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ wp i18n make-mo <source> [<destination>]
     $ wp i18n make-mo .
 
     # Create a MO file from a single PO file in a specific directory.
-    $ wp i18n make-mo example-plugin-de_DE.po languages
+    $ wp i18n make-mo languages/example-plugin-de_DE.po
 
     # Create a MO file from a single PO file to a specific file destination
     $ wp i18n make-mo example-plugin-de_DE.po languages/bar.mo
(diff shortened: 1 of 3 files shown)


swissspidy:
Thanks for opening this. Not sure about the suggested changes 🤔 

`wp make-mo <source> [<destination>]` is definitely supported and covered by our tests.

`wp i18n make-mo woocommerce-de_DE.po wp-content/languages/plugins` definitely works, if there is a `woocommerce-de_DE.po` in the current working directory.

It sounds like you're trying to run the command from a different directory where that file doesn't exist, in which case of course it will fail. And then, yes, `wp i18n make-mo path/to/my/translations/woocommerce-de_DE.po` works. But that's different from `wp i18n make-mo path/to/my/translations/woocommerce-de_DE.po path/to/destination/folder`

swissspidy:
Closing as per the above

sun:
That is not what the documentation suggests. Specifically, to run the command inside of the directory already.

I understand you want to keep the queue clean by closing this, but it’s a bit too quick and premature.

swissspidy:
I‘m happy to accept a docs change with a better example, but the current suggested change js just wrong and doesn‘t match the description.

Issue #435, opened 2025-03-25
Add support for custom translation keys


michaelw85:
## Feature Request

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).

**Describe your use case and the problem you are facing**

When using a custom templating engine like Laravel blade or twig it can be interesting to create your own translation function or even just a custom function in WP. I use blade and created a directive `@t()` the wraps `__()` and adds a bit more functionality like a default text domain and string interpolation. As expected, these custom keys/functions are not included in the resulting pot file.

**Describe the solution you'd like**

I would love to add an optional argument allowing custom keys/functions to be included in the resulting pot.
For example the `-k` flag available in the [xgettext](https://linux.die.net/man/1/xgettext) command.

michaelw85:
I have time to contribute this friday and would like to see if this is something I could add unless this ticket is rejected.

michaelw85:
> Blade is currently supported by default, but of course only with the default `__()` functions. Twig has come up before, see [#52](https://github.com/wp-cli/i18n-command/issues/52).
> 
> I don't see us adding something like the `--keyword` argument in `xgettext`, as it would add a lot of extra complexity and wouldn't even be able to handle anything custom like you describe it.
> 
> I'd suggest looking into extending the base functionality using your own package, similar to how https://github.com/timber/wp-i18n-twig does it. Then you can implement all sorts of extra logic for your custom directives.

Thanks for the feedback, adding my own plugin/package was something I was thinking about but since I started contributing I wanted to check if it could be a core feature.

FYI, I only added `-kt` to my xgettext command to make it pickup my directive `@t('key')`. Only downside of my current setup is that I forced xgettext into python language to support blade files. I want to switch to wp cli to generate the pot files.

PR #445, opened 2025-05-22
Add support for custom translation blade directives

michaelw85:
Solves #435 

Add a new option `--blade-directives` making it possible to use custom directives as translation functions.
These directives will be rewritten to `__(` to allow them to be picked up as translation strings.

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -3092,6 +3092,38 @@ Feature: Generate a POT file of a WordPress project
       msgid "Page not found."
       """
 
+@blade
+  Scenario: Extract strings from custom directives from a Blade-PHP file in a theme
+    Given an empty foo-theme directory
+    And a foo-theme/style.css file:
+      """
+      /*
+      Theme Name:     Foo Theme
+      Theme URI:      https://example.com
+      Description:
+      Author:
+      Author URI:
+      Version:        0.1.0
+      License:        GPL-2.0+
+      Text Domain:    foo-theme
+      */
+      """
+    And a foo-theme/stuff.blade.php file:
+      """
+        @t( 'This is a custom directive', 'foo-theme' )
+      """
+
+    When I try `wp i18n make-pot foo-theme --blade-directives=t result.pot --debug`
+    Then STDOUT should be:
+      """
+      Theme stylesheet detected.
+      Success: POT file successfully generated.
+      """
+    And the result.pot file should contain:
+      """
+      msgid "This is a custom directive"
+      """
+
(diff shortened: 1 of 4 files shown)


swissspidy:
Thanks for your PR.

As per https://github.com/wp-cli/i18n-command/issues/435#issuecomment-2750583101, I am hesitant to add something like this. This change may work for your particular use case, but not for everyone else's.

It's unclear how this would work for other functions like `_x()` or `_n()`. And what if someone has different directives with different signatures?

That's why I said this quickly leads to additional complexity.

michaelw85:
> As per [#435 (comment)](https://github.com/wp-cli/i18n-command/issues/435#issuecomment-2750583101), I am hesitant to add something like this. This change may work for your particular use case, but not for everyone else's.

I thought I would give it a go, and you might accept it since it's an optional argument. I understand that in its current state, it is very limited in what it supports. The only way to fully support this feature would be to inject custom directives, and I don't think this would be possible in any way. This would mix your local WP code with WP-CLI.

> It's unclear how this would work for other functions like `_x()` or `_n()`. And what if someone has different directives with different signatures?

I could update the syntax for the params to something like `@t:__` to support a custom mapping.

If you don't want to accept this change, I would ask if I can achieve this in a custom plugin? I haven't made any yet and I would think I'm making changes to something that I can't extend.

swissspidy:
> If you don't want to accept this change, I would ask if I can achieve this in a custom plugin? I haven't made any yet and I would think I'm making changes to something that I can't extend.

Not in a WordPress plugin, but in a custom WP-CLI command. You have already forked the repository for this PR, so you could install that via `wp package install https://github.com/michaelw85/i18n-command` and then `wp i18n` would be using your customized version.

michaelw85:
Thx for the info! I will close this pr and ticket and will go with a custom install.

michaelw85:
> > If you don't want to accept this change, I would ask if I can achieve this in a custom plugin? I haven't made any yet and I would think I'm making changes to something that I can't extend.
> 
> Not in a WordPress plugin, but in a custom WP-CLI command. You have already forked the repository for this PR, so you could install that via `wp package install https://github.com/michaelw85/i18n-command` and then `wp i18n` would be using your customized version.

FYI installing a custom branch will not completely replace an existing one. Due to how composer works it will still autoload certain classes from the original code so it becomes part custom, part original code 😅 . Fun little challenge I'm currently working on.
For the install I needed to use `wp package install user/repo`, the full url was giving me errors for some reason.

Issue #449, opened 2025-07-25
`.min.js.map` files not excluded in make-pot, causing `.min.js` to be parsed and leading to memory issues


vipul0425:
## Bug Report

- [x] Yes, I reviewed the [contribution guidelines](https://make.wordpress.org/cli/handbook/contributing/).
- [x] Yes, more specifically, I reviewed the guidelines on [how to write clear bug reports](https://make.wordpress.org/cli/handbook/bug-reports/).

**Describe the current, buggy behavior**

The `make-pot` command attempts to parse `.min.js` files even though they are explicitly listed in the excluded patterns `(*.min.js)`. This happens when the corresponding `.min.js.map` files are present in the same directory.

In `--debug mode`, the log misleadingly shows that `.min.js` files are being parsed. However, the actual culprit appears to be the `.min.js.map` files. This can cause unnecessary processing and even memory exhaustion issues in large codebases with many `.min.js` and `.min.js.map` files.

**Describe how other contributors can replicate this bug**

- Create a build directory with a `.min.js` file and its corresponding `.min.js.map` file.
- Run the `wp i18n make-pot` command on the directory with `--debug` enabled.
- Observe the debug output showing `.min.js` being parsed despite being excluded.
- In larger projects, this may lead to memory exhaustion or long processing times.

**Describe what you would expect as the correct outcome**

Ideally, `.min.js.map` files should also be excluded from parsing just like `.min.js`.
At the very least, the debug output should clearly state when `.map` files are being parsed, to aid troubleshooting.

**Provide a possible solution**

Update the exclusion list to also include `*.min.js.map` or `.map`.

**Provide additional context/Screenshots**

<img width="1522" height="55" alt="Image" src="https://github.com/user-attachments/assets/4a93f085-32df-4717-91d4-afd61116845f" />

<img width="489" height="63" alt="Image" src="https://github.com/user-attachments/assets/713068ad-f3cc-4d43-9a6b-9aa5ce36b736" />

<img width="1264" height="87" alt="Image" src="https://github.com/user-attachments/assets/8bb19408-b9e8-4838-a7f5-9c479439cf86" />

swissspidy:
Thanks for your report.

The message is a bit misleading / incorrect. To clarify: `wp i18n make-pot` parses `*.js.map` files **on purpose** to get translations with correct source information. It does not actually parse your `.min.js` file (again, the message is incorrect). We do not want to exclude `.js.map` files.

PR #450, opened 2025-07-25
Fix GH 449: Exclude `*.min.js.map` files from translation in make-pot command

hbhalodia:
- Issue - #449


swissspidy:
Closing as per https://github.com/wp-cli/i18n-command/issues/449#issuecomment-3117064760

Surprised the tests aren't actually failing 🤔

PR #468, opened 2025-12-02
Add border.radiusSizes to theme-i18n.json

t-hamano:
- Related WordPress core ticket: https://core.trac.wordpress.org/ticket/64333
- Related WordPress dev note: https://make.wordpress.org/core/2025/11/12/theme-json-border-radius-presets-support-in-wordpress-6-9/

WordPress 6.9 introduces a new `border.radiusSizes` key in theme.json to define border-radius presets.

I have submitted this PR to make these preset names translatable in the `make-pot` command.

### DIFF
--- a/features/makepot.feature
+++ b/features/makepot.feature
@@ -3696,6 +3696,11 @@ Feature: Generate a POT file of a WordPress project
               "fontSizes": [
                   { "name": "Small", "slug": "small", "size": "13px" }
               ]
+          },
+          "border": {
+            "radiusSizes": [
+              { "slug": "soft", "size": "6px", "name": "Soft corners" }
+            ]
           }
         }
       }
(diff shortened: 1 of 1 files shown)


swissspidy:
Thanks. Please note we have automation in place nowadays to handle this:

https://github.com/wp-cli/i18n-command/blob/94f72ddc4be8919f2cea181ba39cd140dd480d64/.github/workflows/update-schemas.yml#L22-L77

These file updates aren't usually necessary anyway as they are just fallbacks for when we can't fetch the file from SVN for whatever reason.

swissspidy:
Yeah that test will fail until the change is merged in core.

swissspidy:
Sure, why not :) You already wrote them, so can't hurt :)

